PR #4, opened 2026-02-27
Update skills, templates, and tags for ASP spec alignment

lhparker1:
## Summary
- Replace decision `type` field with `tags` to match current ASP spec
- Use domain-specific tags in examples instead of hardcoded method/data/parameter
- Update CLAUDE.md template and claude template
- Update `/prism-run`, `/prism-build`, and other skills for Dagster workflow
- Homogenize UI branding and site configs
- Add hooks to promote prism run
- Point ASP dependency at main (inline-recipes merged)

## Test plan
- [ ] Verify skills render correctly in Claude Code
- [ ] Confirm `asp.yaml` examples validate with updated tag schema
- [ ] Check that prism-build skill works end-to-end


src/prism/dagster/sites.py
```diff
@@ -45,81 +47,56 @@
...
-                "container_flags": [],
-            },
-        },
-        "resource_limits": {
-            "max_nodes": 4,
-            "max_walltime_minutes": 180,
-            "max_concurrent_jobs": 8,
-            "max_node_hours_per_session": 32,
-        },
+        "guidance": """\
```
EiffL: @claude please remove this guidance field from the SITE_DEFAULT. We don't want the agent to actually have to know anything about the underlying compute or about how things are executed. for maximal portability everything should just flow from using the prism cli

EiffL:
@claude please remove this guidance field from the SITE_DEFAULT. We don't want the agent to actually have to know anything about the underlying compute or about how things are executed. for maximal portability everything should just flow from using the prism cli

EiffL:
@claude the tests are failing, can you look at it and try to understand why?

PR #6, opened 2026-03-01
Add tiered Claude Code permission presets to prism init

lhparker1:
## Summary

- Adds a `--permissions` flag to `prism init` with four tiers: **yolo** (everything auto-allowed), **recommended** (Prism workflow auto-allowed), **minimal** (read-only), and **custom** (recommended + edit yourself)
- First-time users without a saved default are prompted interactively to pick a tier and optionally save it to `~/.prism/config.yaml`
- Subsequent inits silently use the saved default; `--permissions` flag overrides without changing the saved default
- Hooks remain identical across all tiers

## Test plan

- [x] Dry-run verified all 4 tiers produce correct `settings.json` permissions
- [x] Interactive prompt flow tested (pick tier, save default)
- [x] Saved-default flow tested (silent reuse on subsequent init)
- [x] Flag override tested (overrides saved default without changing it)
- [x] Hooks present in all tiers confirmed
- [ ] Existing test suite passes with `--permissions recommended` added to all init invocations

### DIFF
--- a/src/prism/cli.py
+++ b/src/prism/cli.py
@@ -16,6 +16,42 @@ from rich.console import Console
 
 console = Console()
 
+PERMISSION_TIERS = {
+    "yolo": {
+        "allow": [
+            "Bash(*)",
+            "Edit",
+            "Read",
+            "Write",
+            "WebSearch",
+            "WebFetch",
+            "mcp__*",
+        ],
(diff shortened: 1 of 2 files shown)


EiffL (approved):
LGTM

PR #7, opened 2026-03-01
Remove /prism skill, fix schema docs

lhparker1:
## Summary

- **Delete the `/prism` skill** — it was confusing since users should never invoke it directly. Its content was ~90% redundant with the CLAUDE.md template. The few unique pieces (routing table, artifact evidence) are relocated to the template.
- **Fix schema documentation** in the CLAUDE.md template: `success_criteria` now uses structured objects, and `when`, `incompatible_with`/`requires`, and `excluded`/`excluded_reason` are documented.
- **Update cross-references** in `prism-new`, `prism-build`, `prism-verify` to point at the project root `CLAUDE.md` instead of the deleted skill.
- **Fix cli.py fallback** string that referenced the deleted skill.

## Test plan

- [ ] Verify `prism init` still produces a correct `CLAUDE.md` with the routing table and schema fixes
- [ ] Verify `/prism-new`, `/prism-build`, `/prism-verify` skills load without broken references
- [ ] Verify `/prism` no longer appears as an invocable skill


EiffL (approved):
LGTM!

PR #13, opened 2026-03-02
Tighten prism-new scoping rules

lhparker1:
## Summary
- **Stop duplicating asp.yaml into CLAUDE.md**: removed Description, Structure, Key Decisions, and Literature Support subsections from the finalize step. CLAUDE.md now only captures Domain Context and Implementation Notes — things that would be lost after `/clear` and aren't in the spec.
- **One output per output**: no bundling multiple metrics or plots into a single output entry.
- **Decisions don't control outputs**: decisions govern *how* something is computed, not *whether* an output exists. Added to the "Not Decisions" section in decision-guide.md.

## Test plan
- [ ] Run `/prism-new` on a test project and verify CLAUDE.md only gets Domain Context and Implementation Notes
- [ ] Verify outputs are created as individual entries during Phase 2
- [ ] Verify decision identification skips "include this output?" candidates


EiffL (approved):
LGTM

PR #47, opened 2026-03-23
Add .langfuse/ and .claude/ to default .gitignore

lhparker1:
## Summary
- Adds `.langfuse/` and `.claude/` to `_GITIGNORE_LINES` so new projects automatically ignore these directories
- Both are created during normal Prism usage but should not be version-controlled

## Test plan
- [ ] Run `prism init` on a fresh project and verify `.gitignore` includes `.langfuse/` and `.claude/`
- [ ] Run `prism init --existing-project` on a project with an existing `.gitignore` and verify the entries are appended

### DIFF
--- a/src/prism/cli.py
+++ b/src/prism/cli.py
@@ -308,6 +308,8 @@ _GITIGNORE_LINES = [
     ".venv/",
     ".ipynb_checkpoints/",
     ".DS_Store",
+    ".langfuse/",
+    ".claude/",
 ]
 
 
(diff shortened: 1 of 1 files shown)


src/prism/cli.py
```diff
@@ -308,6 +308,8 @@ def init(
     ".venv/",
     ".ipynb_checkpoints/",
     ".DS_Store",
+    ".langfuse/",
+    ".claude/",
```
EiffL: @lhparker1 maybe this one we can keep in the git history

EiffL:
```suggestion
```

EiffL (approved):

PR #51, opened 2026-03-25
Feature/sub analysis redesign

lhparker1:


EiffL (commented):
## Sub-Analysis Implementation Review

Focused on the sub-analysis tree logic — ignoring SLURM and ASTRA schema changes.

### 1. Tree walking is one-level deep, but `resolve_analysis_tree` recurses

`collect_tree_outputs` in `tree.py` iterates `spec["analyses"]` but does **not** recurse into nested sub-analyses:

```python
for analysis_id, analysis_node in (spec.get("analyses") or {}).items():
    for out in get_outputs(analysis_node):
        results.append(...)
```

Meanwhile, `resolve_analysis_tree` in ASTRA **does** recurse (`sub_data = resolve_analysis_tree(sub_data, resolved_dir)`). So if someone nests sub-analyses two levels deep, ASTRA resolves them into the spec, but Prism silently drops their outputs, assets, and status.

The same one-level limitation applies to `collect_tree_inputs`, `_display_tree_status`, and `get_output_status`.

**Suggestion:** Either make the tree helpers recursive, or add validation that rejects nested sub-analyses (e.g. error if a resolved sub-analysis itself has `analyses:`). The latter is simpler and probably the right call for now — nesting isn't a stated goal.

### 2. `_init_sub_analysis` doesn't guard against nesting

If you `cd analyses/stage_a && prism init --sub-analysis inner_stage`, it finds `astra.yaml` in the sub-analysis directory and wires into it. This silently creates a nested sub-analysis that Prism's tree helpers can't handle (per issue 1). A check like "if we're already inside a sub-analysis, error out" would prevent confusion.

### 3. Root-level alias outputs with `from:` skip `recipe` — no way to materialize them

In `build_asset_definitions`, alias outputs (root outputs with `from: sub.output`) become bare `AssetSpec`s with no execution logic. Status shows them as `alias`. This is fine as a design choice, but if a user does `prism run final_result`, the asset will be selected but can't be materialized (no `@asset` function). Worth checking that this produces a clear error rather than a silent no-op.

### 4. Stale `when` reference in documentation (fixed in follow-up commit)

`prism-reference.md` line 155 referenced the removed decision `when` syntax. I've pushed a fix for this already.

---

Everything else looks solid — the `TreeOutput` dataclass is a clean abstraction, test coverage is thorough (backward compat tests for flat specs are a nice touch), and the `prism init --sub-analysis` wiring logic is well thought out.

EiffL:
@claude can you fix the issues identified in this review?

PR #61, opened 2026-04-14
Fix 'findings' terminology in prism-extractor agent

EiffL:
## Summary

- Replaced "findings" with "claims" in the prism-extractor agent definition to avoid collision with the ASTRA schema term
- In ASTRA, `findings` specifically means conclusions from the analysis itself (backed by output artifacts), while the extractor produces `prior_insights` (knowledge from literature). Using "findings" to describe what the extractor identifies in papers was ambiguous.
- "Claims" aligns with the `Insight.claim` field name and is unambiguous

## Test plan

- [ ] Verify prism-extractor agent still works correctly with `/prism-new` paper extraction
- [ ] Confirm no other files needed updates (audited all mentions of insights/findings across the repo)


EiffL (commented):
ok

PR #97, opened 2026-04-30
Trim and realign Claude Code skills, hooks, and reference docs

lhparker1:
## Summary
- Realign `lc-new`, `lc-migrate`, `lightcone-cli-reference.md`, and `astra-reference.md` with the current ASTRA spec, scaffold, and CLI surface
- Drop `lc-build`, `lc-verify` skills and `ui-brand.md` guide; consolidate skill set to `lc-new`, `lc-migrate`, `lc-feedback`
- Rewrite `check-lc-run.sh` as pure bash + jq (was importing `lightcone.engine.status` from a Python heredoc, which silently no-op'd in the empty venv `lc init` creates). Add `recipe_command` to `OutputStatus` and `lc status --json` to support the rewrite.
- Fix `session-start.sh` validation preview (was `head -5`, which always landed before the actual error block) and remove a dead `decision_count` line.
- Wire the four project hooks into `.claude/settings.json` on `lc init`; auto-create `~/.lightcone/config.yaml` on first invocation.

## Test plan
- [x] `uv run pytest` — 264 passed
- [x] `uv run ruff check src/ tests/` — clean
- [ ] Manual: `lc init` in a fresh dir, confirm hooks land in `.claude/settings.json` and trigger as expected
- [ ] Manual: edit a malformed `astra.yaml`, confirm `validate-on-save` surfaces real errors
- [ ] Manual: run a recipe script directly, confirm `check-lc-run` warns

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #98, opened 2026-04-30
Trim and realign Claude Code skills, hooks, and reference docs

lhparker1:
## Summary
- Realign `lc-new`, `lc-migrate`, `lightcone-cli-reference.md`, `astra-reference.md`, and the project `CLAUDE.md` template with the current ASTRA spec, scaffold, and CLI surface
- Drop `lc-build`, `lc-verify` skills and `ui-brand.md` guide; consolidate skill set to `lc-new`, `lc-migrate`, `lc-feedback`
- Wire the four project hooks into `.claude/settings.json` on `lc init` (previously bundled but never invoked)
- Rewrite `check-lc-run.sh` as pure bash + jq (was importing `lightcone.engine.status` from a Python heredoc, which silently no-op'd in the empty venv `lc init` creates). Add `recipe_command` to `OutputStatus` and surface it in `lc status --json` so the rewritten hook can match agent invocations against recipe scripts via jq.
- Fix `session-start.sh` validation preview (`head -5` always landed before the real error block), drop a dead `decision_count` line, and remove the lc-build loop block

This PR replaces #97, which had merged #82 history conflicts. This is rebased cleanly on `main` with the same scope.

## Test plan
- [x] `uv run pytest` — 287 passed
- [x] `uv run ruff check src/ tests/` — clean
- [ ] Manual: `lc init` in a fresh dir, confirm hooks land in `.claude/settings.json` and trigger as expected
- [ ] Manual: edit a malformed `astra.yaml`, confirm `validate-on-save` surfaces real errors
- [ ] Manual: run a recipe script directly, confirm `check-lc-run` warns


EiffL:
@claude why do we touch status.py in this PR?

EiffL (approved):
LGTM

PR #104, opened 2026-04-30
Streamline Claude Code project hooks (fixes #103)

### LINKED ISSUE #103: PostToolUse Write hook reports stale schema validation failures
## What happened

The PostToolUse:Write hook and `astra validate` disagree on **schema** validation of the same astra.yaml.

After writing the file, the hook reported:

```
ASTRA validation FAILED for astra.yaml:
Schema validation errors:
  • outputs.0: Additional properties are not allowed ('decisions', 'inputs' were unexpected)
  • outputs.1: Additional properties are not allowed ('inputs' was unexpected)
  • outputs.2: Additional properties are not allowed ('inputs' was unexpected)
```

Running `astra validate astra.yaml` immediately after in the same shell showed:

```
✓ Schema validation passed
✓ Semantic validation passed
✓ Narrative anchors resolved

Narrative section errors:
  • [NARRATIVE_SECTION_REQUIRED] narrative.methods: ...
  • [NARRATIVE_SECTION_REQUIRED] narrative.inputs: ...
  • [NARRATIVE_SECTION_REQUIRED] narrative.outputs: ...
```

The CLI exits 1, but on a *different* set of errors. The schema errors the hook flagged do not appear in the CLI output, and the narrative-section errors the CLI flagged do not appear in the hook output.

Per the current schema (`astra schema show analysis`), `Output` does declare `inputs` and `decisions` as attributes, so the hook's complaint looks wrong against this version of the schema.

## Reproduction

1. Project venv with `astra` 0.2.6 on PATH
2. Write an astra.yaml with `outputs[].inputs` and `outputs[].decisions` (valid per the installed schema)
3. Hook reports `Additional properties are not allowed`; `astra validate astra.yaml` reports schema OK and only flags narrative-section coverage

Likely cause: the hook is resolving a different `astra` binary or schema version than the project venv (or has a stale cached schema).

## Environment

- ASTRA: 0.2.6 (project venv)
- lightcone-cli: 0.3.3
- Python: 3.14.4
- OS: Linux 6.19.11-arch1-1

EiffL:
## Summary

- **Fixes #103**: PostToolUse validator was running an unrelated globally-installed `astra` because `activate-venv.sh` bailed silently under `set -e`. Rewrites it to write `PATH`/`VIRTUAL_ENV` directly to `CLAUDE_ENV_FILE` — the documented mechanism — with no sourcing and no `set -e`.
- **Drops `check-lc-run.sh`**: ~80 lines of awk-parsing shell commands to nudge the agent toward `lc run`. The system already self-corrects via `lc status` showing the output as missing/stale, and CLAUDE.md already says to use `lc run`. The hook was a silent no-op for most projects until recently (it imported a Python module the venv didn't have) — nobody noticed.
- **Slims `session-start.sh`**: drops the model-sync that rewrote `.claude/agents/lc-extractor.md` on every session start (mutating a tracked file at the wrong layer). Switches from Rich-glyph counting to `lc status --json`. Drops project-name / decision-count / universe-count trivia from additionalContext (Claude reads `astra.yaml` and `CLAUDE.md` directly).
- **Slims `validate-on-save.sh`**: drops the binary-resolution fallback branches; trusts PATH from `activate-venv.sh`.
- **Hardcodes `model: sonnet`** in `lc-extractor.md`. Strips `extraction_model` from `lightcone-cli-reference.md` and `lc-new` SKILL.md (config never had this field).
- **Moves hook config out of `commands.py`** into `claude/lightcone/hooks.json` so the wiring lives next to the scripts it references. The CLI now owns only `--permissions` tier selection.

Net: 4 hooks → 3, ~350 LOC of bash → ~140, one runtime side effect on a tracked file removed, one bug class (cross-version `astra` skew) eliminated.

## Test plan

- [x] `uv run ruff check src/ tests/` — clean
- [x] `uv run pytest` — 287/287 pass
- [x] `lc init` smoke test — `.claude/settings.json` produced with the same hook structure as before (minus `check-lc-run`)
- [ ] Open the project in Claude Code, modify `astra.yaml`, confirm `validate-on-save` reports against the project venv's `astra` (i.e. issue #103 doesn't reproduce)
- [ ] Confirm SessionStart `additionalContext` shows validation status + materialization counts + reference pointer


EiffL (commented):
LGTM

PR #105, opened 2026-04-30
Session-scoped Dask scheduler

EiffL:
## Summary

The Dask scheduler now outlives a single ``lc run``: spawned on first use, reused across subsequent runs in the same Claude session, torn down on SessionEnd or after 30 minutes idle.

- **One scheduler per execution context.** The project on a laptop; the SLURM allocation inside ``salloc`` (so ``srun``-spawned workers are reused across every run in the allocation rather than respawned each time).
- **NERSC-correct.** State lives under the existing scratch layout (``<scratch>/.lightcone/dask-scheduler/<key>/``), so flock semantics inherit from the rest of the lock layer.
- **Crash-safe via one primitive.** The daemon holds ``flock`` on ``owner.lock`` for its lifetime; the kernel releases it on clean exit, crash, or SIGKILL. Liveness is probed by trying to take the lock — no PID files, no heartbeats. Idle-timeout (Dask's built-in ``Scheduler.idle_timeout``) is the safety net for the rare case where the SessionEnd hook never fires.

## What changes

- **New** ``src/lightcone/engine/dask_daemon.py`` — ``ensure_scheduler``, ``stop``, and a ``python -m`` daemon entrypoint.
- ``src/lightcone/engine/dask_cluster.py`` shrinks from 200 lines of three branches to a connect-only context manager. The LocalCluster and SLURM-srun branches moved into the daemon, where their lifetime is now session-long.
- The ``DASK_SCHEDULER_ADDRESS`` escape hatch is preserved for CI / external schedulers.
- New ``lc dask stop`` command sends SIGTERM to the running scheduler. The new ``SessionEnd`` hook (``claude/lightcone/scripts/session-end.sh``) calls it for prompt cleanup. Both are best-effort — silent on every "nothing to stop" path.
- ``RunDirs.dask_local`` removed: the daemon owns spill placement (per-key, not per-run) under its own scheduler dir.

## Test plan

- [x] ``uv run pytest`` — all 295 tests pass, including the new ``tests/test_dask_daemon.py`` (key/dirs layout, reuse-vs-spawn, stale-state cleanup, spawn timeout, ``stop`` semantics).
- [x] ``uv run ruff check src/ tests/`` — clean.
- [x] ``uv run mypy src/`` — no new errors (only pre-existing third-party stub gaps).
- [x] End-to-end smoke: spawn → reuse → submit task → stop, verifying ``scheduler.json`` and ``meta.json`` are cleaned up post-shutdown.
- [x] SIGKILL recovery: kill -9 the daemon → next ``ensure_scheduler`` probes the dead address, takes spawn lock, cleans stale files, spawns fresh daemon at a new address.
- [ ] Manual: run two ``lc run`` invocations in one Claude session and confirm the second uses the same scheduler (no second startup latency, no second LocalCluster log line).
- [ ] Manual on Perlmutter (next ``salloc``): confirm the SLURM-keyed scheduler is shared across runs in one allocation, and that SLURM cgroup cleanup tears it down at allocation end.

### DIFF
--- a/claude/lightcone/hooks.json
+++ b/claude/lightcone/hooks.json
@@ -15,6 +15,17 @@
       ]
     }
   ],
+  "SessionEnd": [
+    {
+      "hooks": [
+        {
+          "type": "command",
+          "command": "bash ${CLAUDE_PROJECT_DIR}/.claude/scripts/session-end.sh",
+          "timeout": 10
+        }
+      ]
+    }
+  ],
   "PostToolUse": [
     {
       "matcher": "Write|Edit",
--- a/claude/lightcone/scripts/session-end.sh
+++ b/claude/lightcone/scripts/session-end.sh
@@ -1,0 +1,17 @@
+#!/bin/bash
+# SessionEnd hook: stop the project's session-scoped Dask scheduler.
+#
+# Best-effort and silent. The scheduler self-shuts on idle-timeout
+# (see lightcone.engine.dask_daemon) so failure to fire here only
+# delays cleanup; it does not leak resources indefinitely.
+
+input=$(cat)
+cwd=$(echo "$input" | jq -r '.cwd // empty')
+
+[ -z "$cwd" ] && exit 0
+cd "$cwd" 2>/dev/null || exit 0
+[ -f "astra.yaml" ] || exit 0
+command -v lc &>/dev/null || exit 0
+
+lc dask stop >/dev/null 2>&1 || true
+exit 0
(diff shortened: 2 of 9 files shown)


EiffL:
@claude is this the simplest way to provide the functionality of a persistent Dask cluster when executing locally or within an 'salloc' job? 

Also, we do not want to introduce a command line level mention of Dask, the cli should remain agnostic to the execution backend.

PR #114, opened 2026-05-07
Add workflow to check docs after PR is merged

aboucaud:
This will create if need be an issue with the list of docs to be updated.


EiffL (approved):
LGTM

PR #124, opened 2026-05-08
fix(ci): add dedicated lint workflow and fix mypy errors

aboucaud:
## Summary

- **New `lint.yml` workflow**: runs `ruff check src/ tests/` and `mypy src/` on Python 3.13, in parallel with the test matrix, skipped on draft PRs. Linting was previously absent from CI.
- **`pyproject.toml`**: split `rocrate.*` and `astra.*` into separate mypy overrides — `follow_untyped_imports = true` only on `astra.*` (applying it to `rocrate.*` caused `no-untyped-call` cascades in `wrroc.py`); added override for `daytona_sdk` and `dotenv` which are eval-only deps not installed in the dev group.
- **`commands.py`**: `# type: ignore[misc]` on `astra_init.callback()` — Click types `callback` as `Optional[Callable]` but it is always set at decoration time.
- **`eval/build.py`**: renamed `result` → `dirty` to fix `CompletedProcess[bytes]` assigned to a variable typed as `CompletedProcess[str]`.

Both `ruff check src/ tests/` and `mypy src/` pass cleanly locally before this PR.

## Test plan

- [ ] Confirm the new `Lint` workflow appears and passes on this PR
- [ ] Confirm `ruff` step passes
- [ ] Confirm `mypy` step passes

### DIFF
--- a/src/lightcone/eval/build.py
+++ b/src/lightcone/eval/build.py
@@ -45,11 +45,11 @@ def _get_git_info(repo_root: Path) -> VersionInfo:
         pass
 
     try:
-        result = subprocess.run(
+        dirty = subprocess.run(
             ["git", "diff", "--quiet", "HEAD"],
             capture_output=True, cwd=repo_root,
         )
-        info.lightcone_dirty = result.returncode != 0
+        info.lightcone_dirty = dirty.returncode != 0
     except subprocess.CalledProcessError:
         info.lightcone_dirty = True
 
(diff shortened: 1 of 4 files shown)


src/lightcone/eval/build.py
EiffL, line 1: lol, ok

EiffL (approved):
LGTM

aboucaud:
I have read the Developer Certificate of Origin and I hereby sign the DCO for this PR

PR #125, opened 2026-05-08
Use uv for venv creation in lc init (with python fallback)

### LINKED ISSUE #102: `lc init` dreadfully slow compared to uv install for venv building
The current system will use  pip install to bring the latest version of the lightcone-cli in the .venv during 'lc init'. On my system with very slow internet, it's super slow, like 2mins. Which is very frustrating because otherwise our 'uv install' to install lightcone-cli is like 2s.

aboucaud:
## Summary

- Replaces `python -m venv` + `pip install` with `uv venv --python 3.12` + `uv pip install` in `lc init` for faster environment setup
- Gracefully falls back to the stdlib `venv` + `pip` path when `uv` is not available on the system
- Adds Rich status spinners to keep the user informed during both steps

Closes #102

## Test plan

- [ ] `test_init_venv_uses_uv_when_available` — verifies `uv venv --python 3.12` and `uv pip install` are called when `uv` is on PATH
- [ ] `test_init_venv_falls_back_to_python_when_uv_missing` — verifies `python -m venv` / `pip install` fallback when `uv` is absent

Signed-off-by: Alexandre Boucaud <aboucaud@apc.in2p3.fr>

### DIFF
--- a/src/lightcone/cli/commands.py
+++ b/src/lightcone/cli/commands.py
@@ -222,12 +222,26 @@ def init(
 
     # venv
     if not no_venv:
-        subprocess.run(["python", "-m", "venv", ".venv"], cwd=directory, check=False)
-        subprocess.run(
-            [".venv/bin/python", "-m", "pip", "install", "-q", "lightcone-cli"],
-            cwd=directory,
-            check=False,
-        )
+        if shutil.which("uv"):
+            with console.status("[dim]Creating virtual environment…[/dim]"):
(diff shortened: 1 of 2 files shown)


EiffL:
This looks good @aboucaud ! One question, does this modify the principle of having the venv in .venv? We curretrnyl have a hook that sources the .venv when claude starts. I think this doesn't change things, but just want to check

aboucaud:
No it is transparent

EiffL (approved):
LGTM

aboucaud:
I have read the Developer Certificate of Origin and I hereby sign the DCO for this PR

PR #127, opened 2026-05-11
Add boilerplate for docs versioning

aboucaud:
Inspired on what was done for ASTRA except here I only kept the manual publication and did not port the automated workflow other than the justfile recipes.

Since the docs seem to be handled by Cloudfare, I don't know what kind of setup is required.


EiffL (commented):
LGTM

PR #86, opened 2026-04-30
Paper-reproduction skill bundle in lightcone-cli

cailmdaley:
## Summary

Adds a self-contained **paper-reproduction skill bundle** to `claude/lightcone/skills/`, refreshes the project lifecycle skills with substrate-clarifying renames, and lands a major architectural shift across the bundle (outer ralph loop → persistent orchestrator with named sub-agents → **back to a ralph loop**, with the long middle of a reproduction driven by iterations against a per-paper constitution). `lc init` copies the whole bundle into a project — no plugin marketplace, no separate installs.

### Paper-reproduction bundle (the 6 skills `lc-from-paper` composes)

| Skill | Role |
|---|---|
| [`lc-from-paper`](claude/lightcone/skills/lc-from-paper/SKILL.md) | **Reproduction driver.** ORIENT-first; ORIENT is the single pre-loop phase in the user's main session, running in seven stages — asks for the paper, runs `/paper-extraction` inline to acquire it, interviews the user (grounded in the paper), clones the reference code and runs `/lc-from-code` scan-only (when a repo exists), optionally follows up, then drafts the per-paper `constitution.md` + `CLAUDE.md`. Hands off to a **ralph loop** whose iterations carry the long middle: ARCHITECT → SPECIFY → LITERATURE → IMPLEMENT → RUN → COMPARE. When the loop closes (`constitution.md` `status: closed` after a cold-survey iteration), REVIEW close-out runs back in the user's main session. |
| [`ralph`](claude/lightcone/skills/ralph/SKILL.md) | **Loop substrate.** Authoring (Study → Draft → Refine → Launch the constitution), Launching (the tmux runner at `scripts/ralph`), Loop (Survey → Work → Update → Exit). `lc-from-paper` consumes all three modes; standalone for any other long-running work. Replaces the old `constitution` + `ralph-loops` split: felt's bundle does it as one skill, the same shape fits here. |
| [`paper-extraction`](claude/lightcone/skills/paper-extraction/SKILL.md) | DOI / arXiv ID → standardized `work/reference/` directory: paper substrate (arXiv LaTeX primary, PDF + Docling fallback), structural `index.json` (figures, tables, outline, citations with resolved DOIs), and a stub `astra.yaml` representing the paper. Invoked from `lc-from-paper`'s ORIENT (Stage 2) as the primary acquisition path; also invoked per cited paper in LITERATURE. |
[...]

PR #130, opened 2026-05-11
Repackage claude/lightcone/ as a Claude Code plugin

cailmdaley:
## Summary

- **Add manifests.** `claude/lightcone/.claude-plugin/plugin.json` declares the `lightcone` plugin; `.claude-plugin/marketplace.json` at the repo root declares the `lightcone-cli` marketplace sourced from `./claude/lightcone`. Both ship in the wheel via `force-include`.
- **Move hook config to plugin convention.** `claude/lightcone/hooks.json` → `claude/lightcone/hooks/hooks.json`, with command paths rewritten from `${CLAUDE_PROJECT_DIR}/.claude/scripts/X` to `${CLAUDE_PLUGIN_ROOT}/scripts/X`. Hook *scripts* keep the `scripts/` directory name (rather than renaming to `hooks-handlers/`) — they're already cleanly named and minimum churn won.
- **`lc init` shells out instead of copying.** The old 25-line `shutil.copytree` block becomes `claude plugin marketplace add <root>` + `claude plugin install lightcone@lightcone-cli`. The plugin (skills, agents, hooks) lives user-scoped under `~/.claude/`; the project's `.claude/settings.json` keeps only the permissions tier. When `claude` isn't on PATH, `lc init` prints a manual-install hint and continues — Codex / other-agent users still get a clean scaffold.
- **In-bundle paths switch to `${CLAUDE_PLUGIN_ROOT}`.** Every cross-skill reference that used to assume the bundle was copied to `.claude/skills/` (the ralph launcher; `paper-extraction`'s python script) now goes through `${CLAUDE_PLUGIN_ROOT}/skills/<skill>/...`.

Modeled after [`cailmdaley/felt`'s `cmd/setup.go`](https://github.com/cailmdaley/felt/blob/main/cmd/setup.go) — same shape, in Python instead of Go: idempotent CLI shell-out, marketplace-root resolution with bundled + dev fallback, soft-fail when `claude` is absent.

The bundle directory stays at `claude/lightcone/` rather than moving to a more conventional plugin-root path like `claude-plugin/`. Plugin conventions dictate manifest *location and contents*, not the directory the plugin lives in; keeping the path minimises churn across docs and the `force-include` rule.

## History

This branch was originally based on `add/narrative-skill` (#86). After #86 squash-merged into main at `8d979bb`, the branch was **rebuilt clean on top of main** so the plugin contribution stands alone:

```
26df54c  docs + CLAUDE.md: reflect plugin marketplace install model
6486900  bundle: switch in-skill hardcoded paths to ${CLAUDE_PLUGIN_ROOT}
e40e552  repackage claude/lightcone/ as a Claude Code plugin
8d979bb  Paper-reproduction skill bundle in lightcone-cli (#86)   ← main
```

The 3 commits above main are the entire net change; the pre-rebuild branch's other commits were duplicates of work that's now in main's squash.

## Adjacent cleanup folded in

- **`lc-from-paper/references/specify.md`** — a single straggler `Read .claude/guides/decision-guide.md` pointer caught during the in-bundle path sweep; replaced with an explicit `/astra` invocation.

## Test plan

- [x] All 14 `tests/test_cli.py` tests pass, including 3 new ones:
  - `test_init_writes_settings_with_only_permissions` — pins the new `.claude/settings.json` shape (permissions tier only; no embedded `hooks` block, no copied skills/agents/scripts).
  - `test_init_invokes_claude_plugin_when_available` — asserts `claude plugin marketplace add` + `claude plugin install lightcone@lightcone-cli` are invoked when `claude` is on PATH.
  - `test_init_prints_hint_when_claude_missing` — covers the soft-fail branch.
- [x] Autouse fixture `_no_real_claude_plugin_install` monkeypatches `shutil.which("claude")` to `None` by default so the suite never writes to a real `~/.claude/`.
- [x] `lc init /tmp/...` end-to-end — clean project scaffold, marketplace + plugin installed user-scoped, idempotent on re-run.
- [x] `lc init` soft-fail (PATH stripped of `claude`) — clean exit, manual-install hint printed with the correct marketplace root + plugin ref.
- [x] `python -m build --wheel` — confirmed `.claude-plugin/marketplace.json`, `claude/lightcone/.claude-plugin/plugin.json`, and `claude/lightcone/hooks/hooks.json` all land at the right paths inside the wheel.
- [x] `ruff check` — clean across the touched files.

## Notes

- **Base is now `main`** — #86 has landed.
- `claude plugin install` is *user-scoped*. A user who runs `lc init` once has the plugin globally available; subsequent `lc init` calls (new project) are no-ops for the plugin itself. Project-local stale `.claude/skills/` etc. from pre-plugin `lc init` runs are safe to delete — `docs/user/troubleshooting.md` calls this out.

— Claude on behalf of Cail


src/lightcone/cli/commands.py
```diff
@@ -223,10 +233,13 @@ def init(
...
 
-    # Claude Code plugin bundle
-    plugin_source = get_plugin_source_dir()
-    if plugin_source is not None and plugin_source.exists():
-        _install_claude_plugin(directory, plugin_source, permissions)
+    # Agent integrations: write the project's Claude permission tier, then
+    # shell out to each agent CLI so the shared lightcone bundle (skills,
+    # hooks, Pi extension) lives in the user's global agent config rather than
+    # duplicated into every project.
+    _install_claude_plugin(directory, permissions)
```
EiffL, line 240: not correct, we should keep these skills local

EiffL:
let's hold off from merging this branch as we discuss the proper way to support multi harness support.

PR #15, opened 2026-03-03
Add /prism-feedback skill

lhparker1:
## Summary
- Adds `/prism-feedback` skill that lets beta testers file bug reports mid-session
- Claude summarizes the failure from conversation context, triages to the right repo (ASP/Prism/Canvas), and files a GitHub issue via `gh issue create`
- Shows a draft for user confirmation before filing
- Requires `gh` CLI to be authenticated

## Test plan
- [ ] Run `prism init` on a new project and verify `prism-feedback/SKILL.md` is copied to `.claude/skills/`
- [ ] Invoke `/prism-feedback` in a session and verify it checks `gh auth status`
- [ ] Test with an inline description and without (should prompt)
- [ ] Verify the filed issue lands on the correct repo with expected format

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #161, opened 2026-07-19
De-bundle the Claude Code plugin: skills ship from the agent-skills marketplace (LCR-181)

cailmdaley:
Linear: [LCR-181](https://linear.app/lightcone-research/issue/LCR-181) · Design: [de-bundling design doc](https://linear.app/lightcone-research/document/design-plugin-de-bundling-marketplace-skills-global-cli-fd3b4e70440c)

This PR de-bundles the Claude Code plugin from the wheel. Skills, hooks, and the `lc-extractor` agent now ship from the [`agent-skills`](https://github.com/LightconeResearch/agent-skills) marketplace. The CLI keeps the commands, the scaffold, and the compatibility surface the plugin talks to.

## The install path

The CLI and the plugin are installed globally, once. Projects only activate.

`uv tool install lightcone-cli` installs one global command: `lc`. `lc init` registers the agent-skills marketplace globally with each harness on the PATH: `claude plugin marketplace add` for Claude Code, `codex plugin marketplace add` + `codex plugin add` for Codex. Registration is idempotent and best-effort; a missing harness is skipped silently.

Activation is per project. `lc init` writes one key into `.claude/settings.json`: `enabledPlugins`. The plugin loads only in `lc init` directories, on folder trust. The key carries no version and no marketplace source. Codex has no project-scoped plugin config yet (openai/codex#18115), so its plugin stays globally active.

`astra` is global too. The lightcone-cli wheel exposes both commands, so one `uv tool install` yields one `lc` and one `astra` per machine — no version skew between them. `lc init` still creates the project venv, but empty — it installs nothing into it. The venv is the project's analysis environment: the plugin's session-start hook activates it, so packages the agent installs land there. Containerized analysis dependencies flow through `requirements.txt` into the container. When an analysis needs validation against its exact declared spec version, `uvx --from astra-tools==X --with astra-spec==Y astra` gives a spec-matched toolchain per invocation. 

`lc init` scaffolds new projects only: the target directory must be empty or nonexistent, checked before any file is written. Re-running init and adopting an existing ASTRA project are unsupported — `astra init` hard-fails on non-empty directories, so a convergent init needs an upstream astra-tools change first (design doc, Open Questions).

## The upgrade model

Two artifacts carry two versions:

- The **plugin** ships from the marketplace. A skill edit ships without a CLI release.
- The **CLI** ships from PyPI via uv. A CLI change that alters the plugin contract requires a plugin version bump.

There is one plugin install per machine, so there is one plugin version per machine. When a CLI change alters the plugin contract, the plugin version bumps with it.

Compatibility enforcement lives on the plugin side, not in this PR. The plugin manifest will declare a compatible `lc` version range. A harness-agnostic session-start hook will check it against `lc version`. That work ships from the agent-skills repo (astra plugin: agent-skills#15; the lc-paired compat hooks: stacked agent-skills#17 / LCR-189). Until it lands, the version numbers alone carry the contract. There is no handshake command in the CLI.

## Permissions

`lc init` no longer writes a permission policy. The permission-tier mechanism is removed; the harness owns the trust level. The docs now recommend the harness-native modes: Claude Code's accept-edits and Auto modes, Codex's "Approve for me" (a.k.a. Auto-review). A future best-practices skill could carry this guidance.

## Open questions

- **Is the first marketplace release a breaking release?** No official release exists yet. The answer decides how much back-compat machinery we owe.
- **Migration.** Existing pre-plugin projects, and old plugins that call removed subcommands, are out of scope here. They get their own PR once this design settles. A project init'd during dogfooding keeps a stale `extraKnownMarketplaces` key in its `.claude/settings.json`; leaving it is intentional — cleanup belongs to the migration PR.
- **Version scheme.** Per-project pinning is off the table: the harness tracks the marketplace's current version, and the install is global. The current direction is a plugin-declared `lc` version range (minimum, possibly maximum), checked by a session-start hook. The exact scheme is still open. Opinions welcome.

## Testing

Full test suite green (344 tests, ruff, mypy).

## Merge checklist

- [x] ~~astra-tools pin~~ — astra-tools 0.2.11 is on PyPI (released 2026-07-23); the git direct reference is gone and the dependency is a plain `>=0.2.11` (`8bda862`).
- [ ] **Flip the marketplace ref to the default branch when agent-skills#17 merges** (the stacked lightcone CLI-compat PR; #15 alone does not put the lc-paired hooks on the default branch). `MARKETPLACE_BRANCH` in `cli/commands.py` and the `@astra-plugin-rework` suffix in `eval/sandbox.py` track the agent-skills PR branch on purpose (INTERIM comments mark both). One small commit drops them at merge time.
- [ ] **Tag the merge commit `v0.4.0`.** The agent-skills plugin declares `0.4.0 ≤ lc < 0.5.0` with 0.4.0 = first de-bundled release. Version is VCS-derived; without the tag every session warns on version skew.


EiffL:
Superseeded by #168

PR #163, opened 2026-07-22
JupyterHub on GCP: transparent Cloud Build images + run-scoped Dask Gateway clusters (LCR-174)

EiffL:
## Summary

Implements the [Kubernetes/JupyterHub-on-GCP PRD](https://linear.app/lightcone-research/document/prd-integration-with-kubernetes-and-jupyterhub-on-gcp-29f86d16fcbb) (LCR-174) against the **hub-deploy `dask-cloudbuild` deployment**: on a lightcone hub, `lc run` makes sure the project image is up to date (building through **GCP Cloud Build** when needed), **creates a run-scoped Dask Gateway cluster with that image, runs the pipeline in worker pods, and culls the cluster when the run finishes** — the exact same zero-configuration UX as on a laptop or a SLURM allocation. Everything keys off the env contract the deployment injects into user pods (`DASK_GATEWAY__*`, `LIGHTCONE_REGISTRY`, `LIGHTCONE_BUILD_BUCKET`, `LIGHTCONE_BUILD_SERVICE_ACCOUNT`); nothing is configured by the user.

## Design

**One new runtime, not a new subsystem.** The spec already routes everything through two seams — the container runtime (`wrap_recipe`, image resolution) and the cluster context (`cluster_for_run`) — so the k8s integration is a value flowing through existing seams:

- **`kubernetes` runtime** (`engine/container.py`): the worker pod *is* the container. `wrap_recipe` is a passthrough (no nested containers), and Containerfile specs resolve to registry refs. One content-addressed identity everywhere: `image_identity()` yields the digest that spells `lc-<project>-<hash>` in a local store and `$LIGHTCONE_REGISTRY/lc-<project>:<hash>` in the registry — so `code_version`, staleness, and reruns work identically on every backend.
- **Build backend** (`engine/cloudbuild.py`, ~400 lines): selected when the deployment contract is present. Freshness is a single registry HEAD on the content-addressed ref — unchanged files never rebuild, never even upload. A build tars the **staged build context** (the exact file set the tag hashes), uploads it to the deployment's GCS bucket, and submits a Cloud Build job that runs as the deployment's least-privilege build SA. Auth is the pod's Workload Identity via the metadata server — pure `urllib`, no SDK dependency, no git remote, no stored credentials. Failures surface the build-log tail.
- **Gateway branch of `cluster_for_run`** (`engine/dask_cluster.py`): create a cluster with the project image as the `image` cluster option, scale adaptively `1..--jobs`, wait (bounded, default 600 s) for the first worker so an unpullable image fails loudly instead of hanging at zero workers, and **shut the cluster down on exit — success or failure**. Create/cull per run is PRD decision #1 and is also what makes image updates seamless: a Gateway cluster's image is fixed at creation. A startup assertion refuses clusters whose workers don't advertise the `cpus`/`memory` resource contract (the other silent-hang mode).
- **Self-provisioned worker environment — the deployment stays near-stock**: lc reads the gateway's declared cluster options and passes everything workers need through the *standard* `environment` option — the `DASK_DISTRIBUTED__WORKER__RESOURCES__*` scheduling contract mirrored from the declared `worker_cores`/`worker_memory`, the driver's `HOME`/`USER`/`LOGNAME` (passwd-less uid-1000 images crash `getpass.getuser()` without them), and `LIGHTCONE_WORKER_IMAGE` (covering the deployment-default-image case too), preserving any deployment-declared environment defaults underneath. The hub's options handler reduces to the stock 2i2c daskhub shape (cores/memory/image/environment) plus the per-user home mount — zero lightcone-specific injection server-side (hub-deploy `e10088c`).
- **Executor rendezvous by name**: `gateway://` schedulers can't be dialled by a bare `Client`, so `lc run` hands the executor the cluster name (`LIGHTCONE_GATEWAY_CLUSTER`) and it rejoins through the authenticated Gateway API as a guest (`shutdown_on_close=False`).
- **Output through the future result**: worker pods' stdout goes to pod logs, so `_run_shell` now returns `(exit_code, sentinel_block)` and the driver prints each finished rule's block — the one channel that works uniformly across LocalCluster threads, srun workers, and Gateway pods. This **retires the cross-node stdout `flock`** entirely (one less NFS/DVS concern), and a child snakemake that dies before the rule body (missing package in the worker image) forwards a bounded raw tail instead of vanishing (PRD architecture bullet / LCR-175 semantics). Legacy bare-`int` results from older workers are still accepted (version skew on image-based deployments). The executor also prefixes every spawned job with `cd <workdir>` — spawned snakemake commands carry no `--directory`, and Gateway worker pods start in the image's `WORKDIR`, not the project (local/SLURM workers only worked by inheriting the driver's cwd).
- **Site detection by env markers** (`engine/site_registry.py`): a pod's hostname is noise; `DASK_GATEWAY__ADDRESS` is the signal. The `jupyterhub` site declares `container_runtime: kubernetes` and `scratch_root: $HOME` — **no separate scratch space on the hub**; home is the NFS volume shared with every worker pod, so driver and workers see one consistent snakemake state.
- **One snakemake invocation on every backend**: `--shared-fs-usage` minus `software-deployment` is unconditional, so spawned jobs always run plain `python` from the worker's own environment (the worker image on the gateway; the driver's activated env locally/SLURM) instead of embedding the driver's `sys.executable`. No gateway-specific flags remain (`--latency-wait` dropped — the NFS latency was not reproducible in recent tests).
- **Environment-agnostic scaffold** — `lc init` writes the same Containerfile everywhere; `requirements.txt` pins `lightcone-cli`, which carries the whole execution stack (snakemake, dask, distributed, dask-gateway — now a normal dependency, no `[gateway]` extra). No uid/USER is baked into the image: cluster pods run as the notebook uid via the deployment's `securityContext` (hub-deploy `c8dc6cd`), and `getpass.getuser()` works through the lc-provisioned `USER`/`LOGNAME`/`HOME` (LCR-176 Part A).
- **Provenance**: the manifest keeps recording the declared spec, `code_version` hashes the registry ref (Containerfile edit → new ref → rerun), and a new additive `worker_image` field records `LIGHTCONE_WORKER_IMAGE` — provisioned into every pod by lc itself — as ground truth of what actually executed.

A spec declaring several distinct containers is rejected on this backend with guidance (the worker pod image is fixed per run — LCR-176 note); a project with no containers runs on the deployment's default image.

## Out of scope (per PRD)

- Cluster lifecycle beyond one run (clusters never survive between `lc run` calls).
- Non-GCP build backends; general k8s support beyond the lightcone hub contract.

## Relation to #162

Same PRD, different mechanism: #162 built through the BinderHub service (git-ref-based, requires committed/pushed state and a GitHub connect flow). This PR targets the newer Cloud-Build-based deployment (hub-deploy `dask-cloudbuild`), where builds are git-free tarball submissions — which removes the GitHub coupling and most of the surface area (~1.1k src lines vs ~4k).

## Testing

- `just lint` clean (ruff + strict mypy), **388 tests pass** (+55 new: Cloud Build control flow against mocked HTTP seams incl. failure-tail and force-rebuild, gateway create/cull lifecycle incl. failure-path culling and zero-worker timeout, self-provisioned worker environment, an image-override regression test against the real dask-gateway client (lc's explicit image must beat the ambient `DASK_GATEWAY__CLUSTER__OPTIONS__IMAGE` default), executor rendezvous-by-name / workdir-prefix / legacy-result unpacking, sentinel-block forwarding, env-marker site detection, hub scratch resolution, runtime-aware status resolution, kubernetes-runtime Snakefile generation, environment-agnostic `lc init` scaffold).
- **Live end-to-end validated on the deployed lightcone hub** (project `union2.1_lcdm_fit`, user pod → Cloud Build cached-image check → run-scoped Gateway cluster with the project image → all four rules executed in worker pods with per-rule output on the driver terminal → `lc verify` green, manifests recording `worker_image` ground truth → cluster culled on exit). The live test surfaced and fixed three integration bugs: gateway pods must use the modern `dask scheduler`/`dask worker` subcommands (deployment-side, hub-deploy `bfec4ab`), the executor must `cd` into the workdir for spawned jobs (`c7b18e3`), and the status walker must resolve image identities runtime-aware (`50508b9`). A second live validation (forced full re-run) was done after the near-stock streamlining (`3533b25` + hub-deploy `e10088c`): all rules on workers, `lc status`/`lc verify` green, manifests carrying the lc-provisioned `worker_image`, cluster culled — and the run automatically picked up a changed build context (new content hash → Cloud Build → fresh cluster on the new image).

Refs: LCR-174 (PRD), LCR-176 (Part A + build path).

https://claude.ai/code/session_015wg3CVQ5oYWJBCfq58Z8d2

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -59,6 +59,11 @@ docs = [
 ]
 
 
+[project.optional-dependencies]
+# Dask Gateway client for JupyterHub/Kubernetes deployments. The
+# client's minor version should match the deployment's gateway server.
+gateway = ["dask-gateway>=2024.1"]
+
 [project.scripts]
 lc = "lightcone.cli:main"
 
@@ -100,7 +105,7 @@ explicit_package_bases = true
 mypy_path = "src"
 
 [[tool.mypy.overrides]]
-module = ["dask.*", "distributed.*"]
+module = ["dask.*", "distributed.*", "dask_gateway", "dask_gateway.*"]
 ignore_missing_imports = true
 follow_untyped_imports = true
 
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -53,6 +53,11 @@ dev = [
     "ruff",
     "mypy",
     "types-PyYAML",
+    # Real gateway client in the dev env: the image-override contract
+    # (lc's explicit image kwarg must beat the deployment's ambient
+    # DASK_GATEWAY__CLUSTER__OPTIONS__IMAGE default) is pinned by a
+    # regression test against the actual client merge logic.
+    "dask-gateway>=2024.1",
 ]
 docs = [
     "zensical>=0.0.33",
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -110,9 +110,17 @@ explicit_package_bases = true
 mypy_path = "src"
 
 [[tool.mypy.overrides]]
-module = ["dask.*", "distributed.*", "dask_gateway", "dask_gateway.*"]
+module = ["dask.*", "distributed.*"]
 ignore_missing_imports = true
 follow_untyped_imports = true
+
+[[tool.mypy.overrides]]
+# Untyped and without explicit re-exports; skip so `from dask_gateway
+# import Gateway` doesn't trip strict no_implicit_reexport when the
+# package happens to be installed (it's a dev/extra dependency).
+module = ["dask_gateway", "dask_gateway.*"]
+ignore_missing_imports = true
+follow_imports = "skip"
 
 [[tool.mypy.overrides]]
 module = ["rocrate.*"]
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -36,6 +36,11 @@ dependencies = [
     "snakemake-interface-common>=1.14",
     "dask>=2024.1",
     "distributed>=2024.1",
+    # Dask Gateway client for JupyterHub/Kubernetes deployments. A
+    # normal dependency (not an extra) so `lc run` works out of the box
+    # on a hub and scaffolded project images inherit it via their
+    # lightcone-cli pin.
+    "dask-gateway>=2024.1",
     "rocrate>=0.11",
 ]
 
@@ -53,21 +58,11 @@ dev = [
     "ruff",
     "mypy",
     "types-PyYAML",
-    # Real gateway client in the dev env: the image-override contract
-    # (lc's explicit image kwarg must beat the deployment's ambient
-    # DASK_GATEWAY__CLUSTER__OPTIONS__IMAGE default) is pinned by a
-    # regression test against the actual client merge logic.
-    "dask-gateway>=2024.1",
 ]
 docs = [
     "zensical>=0.0.33",
 ]
 
-
-[project.optional-dependencies]
-# Dask Gateway client for JupyterHub/Kubernetes deployments. The
-# client's minor version should match the deployment's gateway server.
-gateway = ["dask-gateway>=2024.1"]
 
 [project.scripts]
 lc = "lightcone.cli:main"
@@ -116,8 +111,7 @@ follow_untyped_imports = true
 
 [[tool.mypy.overrides]]
 # Untyped and without explicit re-exports; skip so `from dask_gateway
-# import Gateway` doesn't trip strict no_implicit_reexport when the
-# package happens to be installed (it's a dev/extra dependency).
+# import Gateway` doesn't trip strict no_implicit_reexport.
 module = ["dask_gateway", "dask_gateway.*"]
 ignore_missing_imports = true
 follow_imports = "skip"
--- a/src/lightcone/cli/commands.py
+++ b/src/lightcone/cli/commands.py
@@ -182,6 +182,7 @@ def init(
 
     from lightcone.engine.site_registry import detect_current_site
 
+    site = detect_current_site()
     directory = directory.resolve()
 
     if (directory / "astra.yaml").exists():
@@ -206,8 +207,17 @@ def init(
             "container: python:3.12-slim", "container: Containerfile", 1
         )
     )
-    (directory / "Containerfile").write_text(_CONTAINERFILE)
-    (directory / "requirements.txt").write_text(_REQUIREMENTS)
+    # On a Dask Gateway deployment the project image doubles as the
+    # worker pod image, so the scaffold must be worker-capable: the
+    # dask/gateway/snakemake/lightcone stack at the versions the hub
+    # runs, and a uid-1000 user matching the mounted NFS home.
+    on_hub = site.get("backend") == "kubernetes"
+    (directory / "Containerfile").write_text(
+        _CONTAINERFILE_WORKER if on_hub else _CONTAINERFILE
+    )
+    (directory / "requirements.txt").write_text(
+        _REQUIREMENTS + _worker_stack_pins() if on_hub else _REQUIREMENTS
+    )
 
     # Append lightcone-specific entries to the .gitignore astra wrote.
     gitignore_path = directory / ".gitignore"
@@ -294,7 +304,6 @@ def init(
     # state (snakemake metadata, dask spill, cross-node locks). On NERSC
     # this is critical: $HOME and CFS are mounted via DVS (no flock, slow
     # small-file I/O), so lightcone keeps everything on $SCRATCH (Lustre).
-    site = detect_current_site()
     if site:
         scratch_expr = scratch_override or site.get("scratch_root")
         if scratch_expr:
@@ -343,6 +352,47 @@ _REQUIREMENTS = """\
 numpy
 pandas
 """
+
+
+_CONTAINERFILE_WORKER = """\
+FROM python:3.12-slim
+
+WORKDIR /app
+
+COPY requirements.txt .
+RUN pip install --no-cache-dir -r requirements.txt
+
+COPY . .
+
+# This image runs as a Dask Gateway worker pod with the user's NFS home
+# mounted at /home/jovyan. A uid-1000 passwd entry keeps outputs owned
+# by the user and getpass.getuser() (called by snakemake) working.
+RUN useradd --create-home --uid 1000 jovyan
+USER jovyan
+"""
+
+
+def _worker_stack_pins() -> str:
+    """Requirements lines for the Dask Gateway worker stack.
+
+    A Gateway worker must speak the deployment's dask protocol and run
+    the child snakemake that executes each rule, so the project image
+    carries dask/distributed/dask-gateway/snakemake/lightcone-cli. Pins
+    mirror the environment ``lc init`` is running in — the notebook pod
+    matches the hub's gateway version, so pinning to it keeps the
+    worker in lockstep. Unpinned fallback for anything not installed
+    here (or installed as a dev build).
+    """
+    from importlib.metadata import PackageNotFoundError, version
+
+    lines = ["", "# Dask Gateway worker stack (pinned to this hub's versions)"]
+    for dist in ("dask", "distributed", "dask-gateway", "snakemake", "lightcone-cli"):
+        try:
+            v = version(dist)
+        except PackageNotFoundError:
+            v = ""
+        lines.append(f"{dist}=={v}" if v and "dev" not in v else dist)
+    return "\n".join(lines) + "\n"
 
 
 _GITIGNORE_APPEND = """
@@ -537,13 +587,14 @@ def run(
     """Materialize outputs declared in astra.yaml.
 
     Always dispatches through a Dask cluster: a ``LocalCluster`` on a
-    workstation, srun-launched workers inside a SLURM allocation, or an
+    workstation, srun-launched workers inside a SLURM allocation, a
+    run-scoped Dask Gateway cluster on a JupyterHub deployment, or an
     existing scheduler if ``DASK_SCHEDULER_ADDRESS`` is set.
     """
     _abort_on_perlmutter_login()
 
     from lightcone.engine.container import load_runtime
-    from lightcone.engine.dask_cluster import cluster_for_run
+    from lightcone.engine.dask_cluster import cluster_for_run, gateway_branch_active
     from lightcone.engine.scratch import (
         RunLockBusyError,
         acquire_run_lock,
@@ -568,10 +619,25 @@ def run(
         console.print(f"[dim]Scratch root:[/dim] {resolve_scratch_root(project)}")
 
     choice = load_runtime(project_path=project)
-    _ensure_images(project, runtime=choice.runtime)
+    images = _ensure_images(project, runtime=choice.runtime)
     snakefile_path, cfg_path = generate(
         project, universes=universes, runtime=choice.runtime
     )
+
+    # On the Gateway branch the cluster is created with one image — the
+    # worker pod is the container for every rule, so a spec declaring
+    # several distinct containers cannot be honoured per-rule.
+    worker_image: str | None = None
+    if gateway_branch_active():
+        if len(images) > 1:
+            raise click.ClickException(
+                "This deployment runs recipes natively in worker pods, "
+                "which supports one container image per run; astra.yaml "
+                "declares several: " + ", ".join(images) + ". "
+                "Consolidate on a single Containerfile (or one shared "
+                "prebuilt image)."
+            )
+        worker_image = images[0] if images else None
 
     # Provenance guard: when ``runtime: auto`` silently fell back to
     # ``none`` and the spec declares any containers, the recipe will run
@@ -624,6 +690,7 @@ def run(
         targets=targets,
         force=force,
         has_outputs=bool(outputs),
+        gateway=gateway_branch_active(),
     )
 
     # Hold a project-level flock for the duration of the run. Acquiring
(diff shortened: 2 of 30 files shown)


EiffL:
**Streamlining update (`3533b25`), live-validated:** the gateway branch now self-provisions the entire worker environment through the deployment's **standard `environment` cluster option** — the `DASK_DISTRIBUTED__WORKER__RESOURCES__*` scheduling contract (mirrored from the gateway's declared `worker_cores`/`worker_memory`), the driver's `HOME`/`USER`/`LOGNAME`, and `LIGHTCONE_WORKER_IMAGE` (now covering the deployment-default-image case too). The matching hub-deploy change (`e10088c`) reduces the options handler to the stock 2i2c daskhub shape (cores/memory/image/environment) plus the per-user home mount — zero lightcone-specific injection left server-side.

Validated on the deployed hub after rolling the near-stock config: forced full re-run of `union2.1_lcdm_fit` — all rules executed in worker pods, `lc status`/`lc verify` green, manifests carry the lc-provisioned `worker_image`, cluster culled on exit. As a bonus the run picked up a changed build context automatically (new content hash → Cloud Build → fresh cluster on the new image).

https://claude.ai/code/session_015wg3CVQ5oYWJBCfq58Z8d2

src/lightcone/cli/commands.py
```diff
@@ -206,8 +207,17 @@ def init(
...
+    # worker pod image, so the scaffold must be worker-capable: the
+    # dask/gateway/snakemake/lightcone stack at the versions the hub
+    # runs, and a uid-1000 user matching the mounted NFS home.
+    on_hub = site.get("backend") == "kubernetes"
+    (directory / "Containerfile").write_text(
+        _CONTAINERFILE_WORKER if on_hub else _CONTAINERFILE
+    )
+    (directory / "requirements.txt").write_text(
+        _REQUIREMENTS + _worker_stack_pins() if on_hub else _REQUIREMENTS
+    )
```
EiffL: Can we avoid doing this branching? I would like the Containerfile to be completely agnostic to the execution environment. We can instead have lightcone-cli pinned in the requirements, with dask-gateway as a dependency.

src/lightcone/cli/commands.py
```diff
@@ -345,6 +354,47 @@ def init(
...
+RUN pip install --no-cache-dir -r requirements.txt
+
+COPY . .
+
+# This image runs as a Dask Gateway worker pod with the user's NFS home
+# mounted at /home/jovyan. A uid-1000 passwd entry keeps outputs owned
+# by the user and getpass.getuser() (called by snakemake) working.
+RUN useradd --create-home --uid 1000 jovyan
+USER jovyan
+"""
```
EiffL: Is this necessary? can it not be taken care of at the level of the dask gateway configuration? so that images do not need to be adapted specifically for this execution environment?

src/lightcone/cli/commands.py
```diff
@@ -345,6 +354,47 @@ def init(
...
+    lines = ["", "# Dask Gateway worker stack (pinned to this hub's versions)"]
+    for dist in ("dask", "distributed", "dask-gateway", "snakemake", "lightcone-cli"):
+        try:
+            v = version(dist)
+        except PackageNotFoundError:
+            v = ""
+        lines.append(f"{dist}=={v}" if v and "dev" not in v else dist)
+    return "\n".join(lines) + "\n"
+
+
```
EiffL: This might not be necessary if we include the ligthcone-cli as a requirement in requirements.txt

src/lightcone/cli/commands.py
```diff
@@ -712,12 +802,26 @@ def _build_snakemake_cmd(
...
     ``--rerun-triggers`` uses ``nargs=+`` in snakemake's argparse, so without
     an explicit ``--`` separator it greedily consumes the first positional
     target path as an extra trigger value, causing an "invalid choice" error.
+
+    *gateway* adapts the invocation to Dask Gateway worker pods:
+
+    * ``--latency-wait 120`` — NFS home volumes sync slowly between
+      worker pods and the driver (measured up to ~60 s on the dev hub,
+      LCR-177); snakemake's default 5 s output-wait would fail
+      successful rules spuriously.
```
EiffL: In our recent tests we did not see this latency, let's remove this special prescription

src/lightcone/cli/commands.py
```diff
@@ -712,12 +802,26 @@ def _build_snakemake_cmd(
...
+    * ``--shared-fs-usage`` without ``software-deployment`` — with it
+      (the default) the child snakemake command embeds the *driver's*
+      ``sys.executable``, a path that doesn't exist inside the worker
+      image; without it, workers invoke plain ``python`` from their own
+      PATH. Everything else stays shared: the NFS home carries the
+      project, sources, and snakemake persistence.
```
EiffL: Shouldn't this 'shared-fs-usage' without software-deployment be the default? does it break anything in the other execution backends if we have it everywhere?

src/lightcone/cli/commands.py
```diff
@@ -734,6 +838,17 @@ def _build_snakemake_cmd(
...
+        cmd += [
+            "--latency-wait",
+            "120",
+            "--shared-fs-usage",
+            "persistence",
+            "input-output",
+            "sources",
+            "storage-local-copies",
+            "source-cache",
+        ]
```
EiffL: I would prefer it if we didn't have to branch the paths here

src/lightcone/engine/dask_cluster.py
```diff
@@ -84,37 +109,269 @@ def _resources_arg(shape: _NodeShape) -> str:
...
+    address, so ``Gateway()`` needs no arguments here.
+    """
+    try:
+        from dask_gateway import Gateway
+    except ImportError as exc:
+        raise RuntimeError(
+            "A Dask Gateway environment was detected "
+            "(DASK_GATEWAY__ADDRESS is set) but the dask-gateway client "
+            "is not installed. Install it with "
+            "`pip install lightcone-cli[gateway]`."
```
EiffL: this is not actually correct, and let's include dask gateway as a normal dependency of lightcone-cli

src/lightcone/engine/site_registry.py
```diff
@@ -76,6 +77,22 @@
...
+    # scratch stays in ``$HOME`` because that's the NFS volume shared
+    # with the Dask worker pods (there is no site-local scratch).
+    "jupyterhub": {
+        "hostname_patterns": [],
+        "env_markers": ["DASK_GATEWAY__ADDRESS"],
+        "display_name": "JupyterHub (Dask Gateway)",
+        "backend": "kubernetes",
+        "connection": {},
+        "container_runtime": "kubernetes",
+        "scratch_root": "$HOME",
```
EiffL, line 101: why do we need to specify a scratch root if not needed for local?

pyproject.toml
```diff
@@ -53,12 +53,22 @@ dev = [
...
 docs = [
     "zensical>=0.0.33",
 ]
 
 
+[project.optional-dependencies]
+# Dask Gateway client for JupyterHub/Kubernetes deployments. The
+# client's minor version should match the deployment's gateway server.
+gateway = ["dask-gateway>=2024.1"]
+
```
EiffL: Let's make it a default dependency

src/lightcone/cli/commands.py
EiffL: Done in 93a2117 — the on-hub branching is gone: `lc init` writes one Containerfile everywhere, and `requirements.txt` pins `lightcone-cli` (unpinned for dev builds), which now carries the whole execution stack — snakemake, dask, distributed, and dask-gateway as normal dependencies.

src/lightcone/cli/commands.py
EiffL: It isn't necessary at the image level — moved to the deployment in hub-deploy `c8dc6cd`: scheduler/worker containers get `securityContext: runAsUser/runAsGroup: 1000`, so NFS-home writes stay user-owned with no USER baked into the image. `getpass.getuser()` doesn't need a passwd entry because it reads `LOGNAME`/`USER` first, which lc already provisions into the pods via the `environment` cluster option.

src/lightcone/cli/commands.py
EiffL: Correct — removed in 93a2117. The lightcone-cli pin brings dask/distributed/dask-gateway/snakemake at compatible versions, so the per-dist ambient pins are gone.

src/lightcone/cli/commands.py
EiffL: Removed in 93a2117 — no `--latency-wait` anywhere; snakemake's 5 s default applies on all backends.

src/lightcone/cli/commands.py
EiffL: Yes — it's now unconditional (93a2117). I checked snakemake's source: dropping `software-deployment` from `--shared-fs-usage` only changes the spawned job command from embedding the driver's `sys.executable` to invoking plain `python` (plus skipping `--scheduler-solver-path`/`--conda-base-path`, which we don't use). That's equally correct on local and SLURM, where workers inherit the driver's activated environment — the one scenario that changes is running `lc` *outside* an activated env (e.g. a bare pipx shim with no venv on PATH), which the project-venv convention already rules out. Validated with a fresh `lc init` → `lc run` → `lc verify` locally.

src/lightcone/cli/commands.py
EiffL: Gone — `_build_snakemake_cmd` lost its `gateway` parameter entirely; one invocation shape for every backend (93a2117).

src/lightcone/engine/dask_cluster.py
EiffL: Done — dask-gateway is a normal dependency now and the ImportError fallback (with its wrong install hint) is deleted; the gateway branch just imports the client (93a2117).

pyproject.toml
EiffL: Done — moved into `dependencies`; the `[gateway]` extra and the duplicate dev-group entry are removed (93a2117).

src/lightcone/engine/site_registry.py
EiffL: This one is load-bearing on the hub, so I kept it and expanded the code comment (93a2117): without a declared scratch root, resolution falls back to the tempdir — fine on a single machine, but a pod's /tmp is pod-local. The driver symlinks `<project>/.snakemake` into scratch, and that state must live on the NFS home every worker pod mounts; with a /tmp fallback each worker would write its job metadata into its own /tmp (invisible to the driver), so every subsequent run would consider all outputs stale. $HOME *is* the shared filesystem here — unlike HPC sites there's no separate scratch, which is why 'local' can go without one but the hub can't.

EiffL:
Review round addressed in `93a2117` + `a45e432` (with hub-deploy `c8dc6cd`, deployed as release revision 7): dask-gateway is a normal dependency, the scaffold is environment-agnostic (pod uid now comes from the deployment's `securityContext`), and the snakemake invocation is identical on every backend (no `--latency-wait`, `--shared-fs-usage` minus `software-deployment` unconditional).

The first live run on the redeployed hub caught one gap — notebook pods don't export `USER`/`LOGNAME`, and the passwd-less agnostic worker image can't fall back to `getpwuid`, so `a45e432` derives the identity on the driver and always provisions it. **Re-validated live after that fix: `lc run` from a user pod completes on Gateway worker pods with the agnostic image.**

src/lightcone/cli/commands.py
aboucaud: Flagging an interaction with the plugin de-bundling PR (#161) before we resolve this thread — the two PRs make opposite assumptions about the project venv, and this change tightens the coupling.

Making `--shared-fs-usage` exclude `software-deployment` makes the **activated environment** load-bearing for rule execution on the local and SLURM backends, in two places:

- **Driver:** the top-level command is a bare `snakemake` console-script, so `snakemake` must be on PATH.
- **Spawned jobs:** the dask executor now spawns plain `python -m snakemake …`, so `python` on the worker's PATH must be able to import `snakemake` + the dask executor plugin.

Both resolve today only because `lc init` installs `lightcone-cli` (hence the whole snakemake/dask stack) into the project `.venv`, which the session-start hook activates. That's the "project-venv convention rules it out" guarantee cited above — and this change now leans on it for spawned jobs too, not just the driver. The SLURM branch already hard-requires it (`dask_cluster.py` aborts if `dask` isn't on PATH inside the allocation's activated env).

The tension: **#161 (de-bundling)** moves to an *empty* project venv with a global `lc`. If that lands, the local/SLURM path here breaks — the driver's `snakemake` isn't on PATH (a global `uv tool install` exposes only `lc`/`astra`), and spawned `python -m snakemake` hits `ModuleNotFoundError`. The **hub/Gateway path is unaffected**: worker pods get the stack from the image via `requirements.txt` (added by this PR), never from the venv.

So whichever of #163/#161 merges second has to reconcile this. If we do drop the venv install, the options are roughly:

1. Invoke snakemake through lc's own interpreter (`[sys.executable, "-m", "snakemake", …]`) and re-add `software-deployment` so spawned jobs use the same interpreter — clean, but partially reverts this thread and reintroduces a per-backend branch for the gateway (where the driver's interpreter path doesn't exist in the worker image).
2. Keep a stack-bearing env on PATH some other way (a global env exposing `snakemake`/`dask`), leaving the empty *project* venv fine.

Not blocking this PR — just flagging so we resolve this thread with eyes open rather than silently.

---
_Generated by [Claude Code](https://claude.ai/code)_

aboucaud (approved):

PR #157, opened 2026-07-05
lc init: scaffold a template MyST report (LCR-167)

EiffL:
## Summary

`lc init` now lays down a template MyST report next to the spec scaffold:

- **`myst.yml`** — registers the MySTRA plugin from its GitHub release URL (`latest`, with a comment showing how to pin a tag) and the stock `book-theme`.
- **`index.md`** — a TODO-driven report skeleton that references the `astra init` boilerplate elements *by path*: the `{astra}` role and directive on `decisions.example_method`, the `outputs` registry embed, and a commented `{astra:value}` example to enable once `lc run` has materialized results.
- `_build/` added to the scaffolded `.gitignore`; the project `CLAUDE.md` stub and the init next-steps now point at `myst start` for local preview.

## Design note

MyST support is deliberately kept as a **recommended add-on of the lightcone layer**, not part of the base ASTRA package: `astra init` and the spec scaffold are untouched, and the report template lives entirely in `lc init`. We can promote it into the spec scaffold later if that proves useful.

**Note on scope**: This issue does not cover adding the authoring skills for telling claude how to use this new capability, this will come next.

Addresses LCR-167.

Signed-off-by: EiffL <fr.eiffel@gmail.com>

PR #165, opened 2026-07-28
Sync docs and Claude skills with v0.4.0

aboucaud:
v0.4.0 shipped the JupyterHub-on-GCP backend (#163) and the MyST report scaffold (#157) without touching `claude/lightcone/` or most of `docs/`. This brings both back on par with the code.

Everything here was verified behaviorally — scaffolding throwaway projects, running generated Snakefiles, and checking real `--help` output — rather than by reading code alone.

## Skills: instructions that actively broke agent runs

These are pre-existing bugs, not v0.4.0 fallout, and several would cause silent wrong behavior rather than an error:

- **`figure-comparison` and `lc-from-paper/references/compare.md` looked for results in the wrong place.** Both searched for flat files at `results/<universe>/<output>.<ext>`. Outputs actually materialize as *directories* containing the artifact plus `.lightcone-manifest.json` (confirmed by running a generated Snakefile). Every comparison would have failed to locate its file and reported a mismatch.
- **`figure-comparison` read the wrong index.** It keyed off `work/reference/metadata.json`, a Docling intermediate that only exists on the PDF fallback path — on the primary arXiv path it is never created. The canonical index is `index.json`.
- **`references/run.md` invented a status vocabulary and a log file.** It told loop iterations to reason about `pending`/`no_recipe` (real states: `ok`/`stale`/`missing`/`alias`, and it omitted `stale` — the one that matters after a spec edit) and to monitor a per-output `.log` that is never written.
- **`references/architect.md`'s stub `astra.yaml` failed the `astra validate` gate the same file imposes.** Replaced with a version iterated through `astra validate` until clean.
- **`references/implement.md` documented a `--universe` auto-injection convention that does not exist** — a recipe using it raises `Unknown recipe placeholder`.
- **`lc-feedback` always reported the ASTRA version as `n/a`**, because `astra.__version__` doesn't exist on a namespace package. Now uses `astra --version`.
- **`session-start.sh` still advertised the removed ASTRA `narrative` field** in every session's primer.

Skills also gained v0.4.0 coverage: the `kubernetes` runtime and Cloud Build path, Dask Gateway detection and the one-image-per-run constraint, and the MyST report scaffold.

## Docs

- Documented what v0.4.0 added: report scaffold, Gateway branch, pre-flight image builds, COPY/ADD source hashing (a rerun-triggering behavior change), `worker_image`, and the new failure modes in troubleshooting.
- Corrected pages that contradicted the code: `site_registry.md` described a now-central module as orphaned; `runner.md` was a "(removed)" tombstone for the module that forms the body of every generated rule; `architecture.md` claimed `.lightcone/lightcone.yaml` is never read (it is).
- Fixed `lc eval` install instructions, wrong in three places — `pip install lightcone-cli[eval]` doesn't work, since `eval` is a dependency-group, not an extra. Added a `docs/cli/eval.md` page.
- Fixed `docs/cli/init.md`, which was wrong about its own flags and about the permission tiers (`/scratch` and `/pscratch` are `ask`, not `deny`).
- New `docs/api/scratch.md` and `docs/api/wrroc.md`; `contributing/{backends,hpc-sites}.md` corrected and linked into nav.
- Deleted the orphaned `docs/hpc/` section: unreachable from nav, redundant with `user/cluster.md`, and it told readers to run `lc setup`, a removed command.
- Adds `CHANGELOG.md`.

## Verification

389 tests pass, ruff clean, `just docs` reports no issues, every nav entry resolves, and the scaffold tree documented in `init.md` matches a fresh `lc init` exactly.

## Known issues found but deliberately not fixed

Code-level, out of scope for a docs/skills sync — worth separate issues:

1. `claude/lightcone/templates/CLAUDE.md` is dead. `lc init` writes an inline `_PROJECT_CLAUDE_MD`; the template is still copied to `.claude/templates/`, so projects carry two divergent versions.
2. The scaffolded `index.md` is coupled to the boilerplate ids `example_method`/`main_result`. Every authoring skill reshapes the spec and breaks those references, and `lc-new` is forbidden from fixing it by its own `allowed-tools`. Skills now warn the user instead.
3. `justfile` eval recipes are broken (`lc eval run` takes a positional config path, has no `--skill` flag, and `--extra eval` targets a dependency-group).
4. `extract_path_b` passes `arxiv_id=None, doi=None` when writing the stub `astra.yaml`, so `--doi` is silently dropped on the PDF path. Looks like a real bug.
5. `lc run` fails on ARM Macs — snakemake's ILP scheduler shells out to pulp's bundled x86-64 CBC binary (`Bad CPU type in executable`). `--scheduler greedy` works.
6. Five `docs/skills/*.md` pages document an "Allowed tools" section for skills that declare no `allowed-tools` frontmatter — fabricated, never drifted. Needs a decision either way.
7. `manifest.SCHEMA_VERSION` is still 1 despite `worker_image` and `git_remote` being added.
8. No test coverage for the plugin bundle — nothing would have caught any of the above.

## One judgment call worth review

`review.md` now asks the user to run `/check-sentence-by-sentence` rather than invoking it, because that skill's description states other skills must not invoke it. Relaxing the skill's description instead would be an equally valid resolution.


claude/lightcone/skills/lc-from-paper/references/review.md
```diff
@@ -32,19 +32,19 @@ The phase name **REVIEW** is freed by the old pre-implement REVIEW phase folding
...
+**Do not spawn `/figure-comparison` under the `Task` tool or inside a ralph iteration.** It prompts the user for inputs it can't infer; sub-agents and detached iterations have no user-reach, so the prompt fires into nothing.
 
 ### `/check-sentence-by-sentence` (opt-in)
 
 Ask the user via `AskUserQuestion` whether they want the claim audit. It's optional because for many reproductions the figure-comparison already settles "did it match?"; the sentence-by-sentence audit earns its keep when the paper makes many specific quantitative claims and the user wants each one anchored to a code location.
 
-If yes, invoke `/check-sentence-by-sentence`. Same discipline as `/figure-comparison` — it can prompt the user; do not spawn under `Task` or inside a ralph iteration.
+If yes, ask the user to run `/check-sentence-by-sentence` themselves — that skill is user-invoked by design and should not be launched from here. Same discipline as `/figure-comparison` once it runs: it can prompt the user, so it belongs in the main session, never under `Task` or inside a ralph iteration.
```
aboucaud, line 45: @cailmdaley what's your view on this ?

claude/lightcone/skills/astra/SKILL.md
```diff
@@ -156,7 +156,7 @@ outputs:
       resources: { cpus: 4, memory: "32GB", gpus: 1, time_limit: "2h" }
 ```
 
-Set `container:` at analysis level (all recipes inherit); per-recipe `container:` overrides. Pass either a container image name (e.g., `python:3.12-slim`, `ghcr.io/org/img:latest`) or a path to a Containerfile (e.g., `Containerfile`, `containers/Dockerfile`). The runtime figures out whether to pull or build.
+Set `container:` at analysis level (all recipes inherit); per-recipe `container:` overrides. Pass either a container image name (e.g., `python:3.12-slim`, `ghcr.io/org/img:latest`) or a path to a Containerfile (e.g., `Containerfile`, `containers/Dockerfile`). The runtime figures out whether to pull or build. Note: on a JupyterHub/Dask-Gateway deployment, `lc run` requires the whole run to resolve to a single container image — a spec declaring several distinct images is rejected there, so consolidate on one Containerfile (or one shared prebuilt image).
```
cailmdaley: not sure if this is critical info

claude/lightcone/skills/lc-from-paper/references/review.md
cailmdaley: this is also said just above on L43, i think it's over-repetitive but this whole skill needs to be reworked so i think it's fine for now

claude/lightcone/skills/lc-new/SKILL.md
```diff
@@ -70,7 +70,7 @@ Spawn all in a single message (parallel). Show progress as results come in:
   ○ Wu & He 2018 (reading...)
 ```
 
-Write extracted prior insights to astra.yaml immediately. Synthesize them by topic for the user.
+Write extracted prior insights to astra.yaml immediately. The extractor's `decision_links:` block is not an ASTRA field — translate it into `insights: [...]` lists on the corresponding decision options when merging (Option.insights in `/astra`); never copy `decision_links` into astra.yaml. Synthesize the insights by topic for the user.
```
cailmdaley, line 73: this is a band-aid, but probably we don't have time to rework the extractor agent. so probably fine for now.

claude/lightcone/skills/lc-new/SKILL.md
```diff
@@ -144,7 +144,7 @@ Show summary table:
 | sub_analysis  | ...       | ...     | ...      |
 ```
 
-Then tell the user the spec is ready and they can begin implementation. Recommend running `/clear` first — the scoping conversation consumes significant context, and everything needed to continue is captured in `astra.yaml` and `CLAUDE.md`.
+Then tell the user the spec is ready and they can begin implementation. Also flag that the scaffolded MyST report (`index.md`) still references the boilerplate ids (`example_method`, `main_result`) and needs updating to the new spec's element paths — the user or a follow-up session can edit it. Recommend running `/clear` first — the scoping conversation consumes significant context, and everything needed to continue is captured in `astra.yaml` and `CLAUDE.md`.
```
cailmdaley: this is not going to be a pleasant experience for users, getting a bug report when they start a new project

cailmdaley (commented):
i think this is an improvement on the existing state, although i'm still concerned that it's not quite up to our standard of quality. some nits, the final one is the most major

claude/lightcone/skills/lc-new/SKILL.md
aboucaud: well noted, will remove this

CHANGELOG.md
EiffL, line 1: let's not include a changelog, the Github release changelog is the cannonical one

EiffL (changes requested):
almost good for me

CHANGELOG.md
EiffL: Done in 24914ee — CHANGELOG.md removed; nothing else in the PR referenced it.

EiffL (approved):
LGTM

PR #173, opened 2026-08-18
Layer 5: the exec boundary and `lc run`

EiffL:
Lands the sandbox layer and the one verb that exercises it.

Per spec §4, `lc run CMD` is the **probe** verb: byte-for-byte the recipe environment — same lock, same `.venv` — under the same boundary recipes will get. A probe has no output, so it needs neither manifests (layer 2) nor the fabric (layer 4). That is what let layer 5 land ahead of them, and it makes it the smallest honest consumer of the boundary that layer 4 will plug recipes into.

No new dependencies.

## The abstraction

**The seam is a pure argv rewrite.** Every mechanism is `Backend.wrap(policy, argv) -> argv'` — a function turning a command into *a different command that sandboxes itself*.

```
                    lc run CMD
                         │
              engine/run.py  ── mode check · rename guard · uv hop
                         │
              sandbox.run(backend, policy, argv, prefix=uv…)
                         │
              backend = detect()          ← the ONLY sys.platform branch
              argv    = backend.wrap(…)   ← pure
              Popen(…)                    ← mechanism-blind
                         │
     ┌───────────────────┼───────────────────┐
 LandlockBackend    SeatbeltBackend      Unavailable
 [python, -m        [sandbox-exec,       argv unchanged
  lightcone.         -p <sbpl>,
  _sandbox_exec,     -DREAD_0=…, --,
  --policy <json>,   *argv]
  --, *argv]
```

Seatbelt is natively that shape; **Landlock is not** — it is a self-restriction, which is precisely what `lightcone/_sandbox_exec.py` exists to fix. Once both reduce to argv, nothing above the seam branches on platform, and **every backend is testable on a host that cannot run it**: `tests/test_sandbox_wrap.py` checks the Landlock wrap *and* the generated macOS SBPL on Linux CI, with no privileges and nothing spawned.

Three types stay deliberately distinct — `Policy` (what we will enforce), `Capability` (what this host can do), `Attestation` (what was actually enforced, spec §7's manifest field). Collapsing any two is how a sandbox starts lying.

Prior art surveyed: OpenAI's codex CLI and this design's own earlier implementation on `redesign_prototype`. They agree on the shape and disagree on one mechanism; this takes codex's answer.

## Deviations from the spec, all recorded in CLAUDE.md

- **The Landlock policy travels as JSON on argv, not as an inherited ruleset FD.** §7 specifies building the ruleset before fork and passing it with `pass_fds`. The shim builds and applies it itself instead — codex's shape. This makes `wrap()` pure and closes §11's own *blocking* spike ("does the Landlock FD survive `uv run`'s spawn/exec chain? an FD cannot be reopened") by making the question moot. It also survives into a container later, where a host FD cannot.
- **Network is not controlled on either platform**, by decision. The generated SBPL explicitly allows it and both platforms attest `network: allowed` — rather than §7's matrix value of `denied`. Nothing pretends to a control it does not apply.
- **`--require-sandbox` is a bare flag.** §7's `=declared-fs` form would be the same flag today, since every existing mechanism scopes the filesystem. It arrives with the first one that doesn't.
- **The denial's system-layer remedy is trimmed** to `uv add` + the ASTRA input declaration. §7's `[tool.lightcone.image]` TOML arrives with the container hatch; printing it now would be a copy-pasteable fix that does nothing.
- **No manifest is written** — a probe has no output. `Attestation.to_manifest()` exists and is tested; layer 4 is its first consumer.
- **One tightening §7 does not contemplate:** `/tmp` leaves the write scope when the project lives under it. Granting it unconditionally would make such a project's tree writable and silently void the read-only-tree guarantee. `TMPDIR` points into the private scope regardless, so `tempfile` works either way.

## Three grants that turned out to be load-bearing

Each found by a real failure while building this, each now pinned by a test:

| Grant | Without it |
|---|---|
| the realpath'd **ELF loader** in the exec set | *every* dynamically linked binary fails EACCES — bash and python included |
| **read** on the uv-managed interpreter's install root | `Failed to import encodings module` — the stdlib sits beside the binary, outside the project and outside `/usr` |
| **read** on `/dev/urandom` | `failed to get random numbers` — CPython seeds hash randomization before `main` |

## Verification

`170 passed`, ruff clean, mypy strict clean. The suite splits along the seam:

- **Pure, every OS** — policy construction, both backends' argv, the golden SBPL, the denial renderer.
- **Real subprocess, no kernel needed** — the shim's contract: its argv, and the reserved exit code `97` that means "lc's own setup broke" and never "the command ran unsandboxed".
- **Real enforcement, no fakes** — `test_sandbox_landlock.py`, module-level `skipif` on `abi() == 0`. A mocked sandbox proves nothing, so where Landlock is absent the file skips whole. Policies there are hand-built so each test grants exactly what it is testing.
- **Pre-exec decisions** — discovery, the containerized refusal, the rename guard, declared inputs, the uv hop.

Exercised by hand end to end against a real scaffolded project (kernel 7.1, Landlock ABI 9): runs in the project `.venv`; project reads allowed; in-tree writes denied *with the file verifiably unchanged*; declared `astra.yaml` inputs become readable; undeclared tools and files denied with their respective remedies; `$HOME` redirected to a private scope; no in-tree `__pycache__`; `--sandbox-debug`, `--no-sandbox`, `--require-sandbox`, and the rename guard all behaving.

https://claude.ai/code/session_01DqTza2GqRLZ4yY5ywAusQH


src/lightcone/cli/commands.py
```diff
@@ -152,3 +156,113 @@ def _render_init_output(report: ConvergenceReport, directory: Path, *, dry_run:
...
+    flag_value=False,
+    default=True,
+    help="Run without enforcement. Recorded as unsandboxed — never silent.",
+)
+@click.option(
+    "--sandbox-debug",
+    "debug",
+    is_flag=True,
+    help="Print the policy and the exact command, then run it.",
+)
```
EiffL: Does this really do something? Also let's remove the option to run without sandbox. so none of these 3 sandbox related things.

EiffL: simplify the rest of the run command accordingly, assuming there is no hatch to escape the sandbox

src/lightcone/cli/commands.py
EiffL, line 1: do we really need these echo and exit status commands? try to streamline and consolidate.

src/lightcone/engine/sandbox/profiles/base.sbpl
```diff
@@ -0,0 +1,134 @@
+; ---------------------------------------------------------------------------
+; VENDORED from openai/codex — codex-rs/sandboxing/src/seatbelt_base_policy.sbpl
+; upstream commit f5e9d66851 (retrieved 2026-08-18)
+;
+; Kept as close to verbatim as possible so it can be diffed against upstream
+; when re-syncing. Local deltas are marked `LIGHTCONE DELTA`.
+; ---------------------------------------------------------------------------
```
EiffL, line 12: Can you reframe is at inspired from...

src/lightcone/engine/sandbox/profiles/platform-defaults.sbpl
```diff
@@ -0,0 +1,206 @@
+; ---------------------------------------------------------------------------
+; VENDORED from openai/codex — codex-rs/sandboxing/src/restricted_read_only_platform_defaults.sbpl
+; upstream commit f5e9d66851 (retrieved 2026-08-18)
+;
```
EiffL, line 4: same comment, inspired from

src/lightcone/engine/sandbox/__init__.py
```diff
@@ -0,0 +1,44 @@
...
+turns that into an argv rewrite — Landlock through
+:mod:`lightcone._sandbox_exec`, Seatbelt through ``sandbox-exec``, and
+:class:`~boundary.Unavailable` through no change at all. :mod:`boundary`
+picks one, runs it, and reports what it enforced.
+
+The threat model is stated, not implied: this enforces
+declared-dependency *discipline* against accidental leakage. It is not a
+boundary against a hostile command — metadata stays visible, an
+interpreter can still read a script it was handed, and memfd-exec is
+unaddressed. All three are adversarial-only.
```
EiffL:
```suggestion
Guarantees that a command cannot use executables or files outside the 
declared set wherever a mechanism exists, and what actually enforced it 
is always recorded.

This includes three components. :mod:`policy` decides *what* is allowed, in
plain path sets that mention no mechanism. A :class:`~model.Backend`
turns that into an argv rewrite — Landlock through
:mod:`lightcone._sandbox_exec`, Seatbelt through ``sandbox-exec``, and
:class:`~boundary.Unavailable` through no change at all. :mod:`boundary`
picks one, runs it, and reports what it enforced.
```

src/lightcone/engine/templates/files/index.md.tmpl
EiffL, line 1: undo the edits to this file, it's fine

src/lightcone/engine/templates/__init__.py
```diff
@@ -56,9 +56,10 @@ def pyproject(*, name: str) -> str:
     """The scaffolded ``pyproject.toml`` for a project called *name*.
 
     Deliberately **virtual** — no ``[build-system]``. Containerized mode
-    builds the image ``--no-install-project`` (project code never enters
-    an image), so a packaged project's ``import my_analysis`` would fail
-    inside its own container (spec §1).
+    will build the image ``--no-install-project`` (project code never
+    enters an image), so a packaged project's ``import my_analysis``
+    would fail inside its own container — which is why the scaffold does
+    not create one (spec §1; the mode itself arrives with layer 6).
```
EiffL: let's remove this paragraph entirely

src/lightcone/engine/templates/__init__.py
```diff
@@ -75,8 +76,8 @@ def python_version() -> str:
     Deliberately not an engine constant: a new project pins the python the
     researcher actually has, rather than one lc would have to download to
     honor a number baked into a release. Identity follows this file from
-    here on — ``env_version`` hashes its bytes, not anything in the engine
-    (spec §3) — so a project is free to repin it afterwards.
+    here on: ``env_version`` will hash its bytes rather than anything in
+    the engine (spec §3, layer 2), so a project is free to repin it.
```
EiffL: never mention the spec in the code and comments

src/lightcone/engine/project.py
```diff
@@ -228,7 +228,13 @@ def converge(directory: Path, *, write: bool = True) -> ConvergenceReport:
     return c.report
 
 
-def _require_uv() -> None:
+def require_uv() -> None:
+    """Refuse early when uv is absent — it is the environment substrate.
+
+    Shared with ``lc run``: convergence needs uv to build the environment
+    and the probe needs it to enter one, and they must not drift into
+    telling the user two different things about the same missing tool.
```
EiffL: remove this 
```suggestion
```

src/lightcone/engine/project.py
```diff
@@ -297,6 +303,51 @@ def project_name(directory: Path) -> str:
...
+    if (directory / "Containerfile.extra").exists():
+        return True
+    pyproject = directory / "pyproject.toml"
+    if not pyproject.exists():
+        return False
+    try:
+        data = tomllib.loads(pyproject.read_text())
+    except tomllib.TOMLDecodeError as e:
+        raise ProjectError(f"{pyproject} is not valid TOML: {e}") from e
+    return "image" in data.get("tool", {}).get("lightcone", {})
```
EiffL: this should NOT be part of this PR, we will only look into system layers later, remove

src/lightcone/engine/project.py
```diff
@@ -297,6 +303,51 @@ def project_name(directory: Path) -> str:
...
+    """
+    start = (start or Path.cwd()).resolve()
+    for directory in [start, *start.parents]:
+        if (directory / SPEC_FILENAME).exists():
+            return directory
+    raise ProjectError(
+        f"no {SPEC_FILENAME} in {start} or any parent directory — "
+        f"not inside an ASTRA project. Create one with `lc init`."
+    )
+
```
EiffL: Let's not do this! 'lc init' and 'lc run' are just going to use their current directory as the root. if the current directory is not a lightcone project, then error, but don't try to find the root of the project otherwise

src/lightcone/engine/run.py
```diff
@@ -0,0 +1,222 @@
...
+is denied, the recipe would have been.
+
+A probe has no output, which is what makes it the strictest consumer of
+the boundary: it may read the project and the declared inputs, and write
+only its own tmp scope. Nothing it does can land in the tree.
+
+The command runs *inside* ``uv run``, and the sandbox wraps the command
+rather than uv: uv, its config, and its caches are trusted plumbing
+outside the boundary (spec §7).
+"""
```
EiffL:
```suggestion
"""``lc run`` executes a command inside the reproducible environment.

Byte-for-byte the environment recipes will get — the same lock, the same
converged ``.venv`` — under the same sandbox. That equivalence is the
point: if a probe works, the recipe will, and if a probe
is denied, the recipe would have been.

A probe has no output, which is what makes it the strictest consumer of
the boundary: it may read the project and the declared inputs, and write
only its own tmp scope. Nothing it does can land in the project tree.
"""
```

src/lightcone/engine/run.py
```diff
@@ -0,0 +1,222 @@
...
+    direct-mode run: the two execute in different worlds and would attest
+    different things.
+    """
+    if declares_system_layer(project):
+        raise ProjectError(
+            "this project declares a system layer, which puts it in "
+            "containerized mode — not available in this release. Remove "
+            "`[tool.lightcone.image]` (and `Containerfile.extra`) to run "
+            "in direct mode."
+        )
```
EiffL: This is future stuff, remove this for now

src/lightcone/engine/run.py
```diff
@@ -0,0 +1,222 @@
...
+    the fabric layer, and until then the message says what it can.
+    """
+    from astra.helpers import get_output_ids
+
+    if not command or command[0] not in get_output_ids(spec):
+        return
+    raise ProjectError(
+        f"`{command[0]}` is a declared output, and `lc run` runs commands, "
+        "not outputs. Materializing outputs arrives with the fabric layer."
+    )
```
EiffL: Let's remove this, we are not guaranteeing that code should be backward friendly/compatible

src/lightcone/engine/run.py
```diff
@@ -0,0 +1,222 @@
...
+    Spec §7 also gives this a ``=declared-fs`` form, which would
+    additionally require a scoped filesystem. Every mechanism that exists
+    today scopes it, so the two forms would be the same flag — it
+    arrives with the first mechanism that does not (the container hatch,
+    where a pod bounds only the OS layer).
+    """
+    if not require or backend.capability.kind != "none":
+        return
+    detail = f" — {backend.capability.detail}" if backend.capability.detail else ""
+    raise ProjectError(f"--require-sandbox: no sandbox mechanism available on this host{detail}")
```
EiffL: this should go since enforcement happens all the time

src/lightcone/engine/run.py
```diff
@@ -0,0 +1,222 @@
...
+
+
+def describe(
```
EiffL: I don't think this sandbox related stuff should be in the 'run.py' module, remove it completely.

src/lightcone/_sandbox_exec.py
```diff
@@ -0,0 +1,326 @@
...
+Two properties of this module are load-bearing and pinned by tests:
+
+- **Stdlib only, and no lightcone imports.** ``lightcone`` is a PEP 420
+  namespace package with no ``__init__``, so ``-m lightcone._sandbox_exec``
+  executes this file and nothing else. It runs on every sandboxed exec;
+  it must not drag the engine in.
+- **It never proceeds unsandboxed.** Every setup failure exits
+  :data:`SETUP_FAILURE_EXIT` with a distinguishable message, so a broken
+  sandbox can never be mistaken for a working one (the "silent
+  best-effort" trap this design exists to avoid).
```
EiffL, line 27: Do not mention the spec and paragraphs in it, never, clean up this comment

EiffL (commented):
Left a few comments

src/lightcone/cli/commands.py
```diff
@@ -152,3 +151,45 @@ def _render_init_output(report: ConvergenceReport, directory: Path, *, dry_run:
...
+    from lightcone.engine.project import current_project
+
+    project = current_project()
+
+    if not command:
+        # On stderr, like every other line lc says about a run: the
+        # command's stdout is the command's. This is the only signal
+        # before an interactive shell takes the terminal, since the
+        # boundary's own notes do not print until the shell exits.
+        click.echo("opening a shell inside the recipe environment", err=True)
```
EiffL: I don't think we should open a shell. what if claude code does 'lc run' then it would open a shell and get stuck because waiting for inputs. Unless you verify that claude code, codex, and open code are all capable of handling shells like this? but I doubt it

src/lightcone/engine/sandbox/profiles/platform-defaults.sbpl
EiffL: maybe include minimal license statement here?

src/lightcone/engine/sandbox/profiles/base.sbpl
EiffL: maybe include minimal license statement here?

tests/conftest.py
```diff
@@ -48,7 +49,20 @@ def fake_run(argv: list[str], *, cwd: Path) -> MagicMock:
...
+    # Linux and does not on macOS, so the enforcement suite was testing
+    # a policy no user would ever get. Fake only what convergence needs.
+    real_which = shutil.which
+
+    def fake_which(name: str, path: str | None = None) -> str | None:
+        if name in ("uv", "git"):
+            return f"/usr/bin/{name}"
+        return real_which(name, path=path)
+
+    monkeypatch.setattr(project.shutil, "which", fake_which)
```
EiffL: hold on.... does that mean that you are adding a 'which' that doesnt exist natively on macos? that's not good.

README.md
```diff
@@ -37,6 +37,33 @@ drafted with any AI coding assistant.
...
+  directory is the project root; erroring beats walking up, guessing,
+  or auto-discovering.
+- **Streamline before shipping.** No small helper functions or
+  rendering layers where a few inline lines read fine; consolidate.
+- **Be honest about provenance.** Third-party material we adapt is
+  "inspired from" upstream, clearly marked — never passed off as
+  verbatim, never left unattributed.
+- **Leave working files alone.** Don't edit files that are fine just
+  because a change nearby made them look touchable.
+
```
EiffL: remove this from the readme

EiffL (commented):
some comments

EiffL (commented):
LGTM

PR #174, opened 2026-08-18
Make the project tree writable, and fix the outside-a-project message

EiffL:
Two small changes to `lc run`, on top of layer 5.

## The project tree is writable now

The boundary is meant to be a no-install stand-in for running the project
in a container. What it should catch is a command reaching **outside** the
project for a tool, library, or data file the image would not have.
Forbidding writes *inside* the project served none of that — a container
bind-mounts the working tree read-write — so the project joins the write
set.

This cannot be narrowed to "everything but `.venv`": Landlock unions
rights over ancestors and has no way to subtract, so a writable project is
necessarily a writable `.venv`. What keeps the environment honest is
`uv sync --exact` at the next convergence, not the sandbox.

Two things existed *only* to serve the read-only tree and go with it:

- the filter that dropped `/tmp` from the write scope when the project
  lived under it;
- `PYTHONPYCACHEPREFIX`, which redirected bytecode into a `$HOME` deleted
  after every run, so every in-tree module recompiled every time. A
  writable tree caches in place — what a container does, and what the
  scaffolded `.gitignore` already expects.

The denial for a readable-but-not-writable path now means *a write outside
the project* (a system path, or a declared input) and says so.

Verified by hand against a real project: writing `results/out.csv` works;
reading an undeclared file under `$HOME`, running `git`, and writing
`/etc/...` are all still denied with their remedies.

### One test note worth reading

The enforcement suite's `outside` fixture moves from `tmp_path` to
`$HOME`. `/tmp` is in the write baseline, so anything `tmp_path` hands us
is **granted** — those denial tests had been passing only because a
project under `/tmp` used to drop `/tmp` from the policy. Left alone they
would have gone green while testing nothing.

## `lc run` outside a project

`/home/francois is missing pyproject.toml, uv.lock, .venv — not a project
root. Run \`lc init\` to converge one.` listed files nobody expected to be
there and advised scaffolding a project in your home directory.

The two failures are different mistakes and now get different advice. No
project markers at all is the wrong *place*: it says so and says to `cd`
to the root of one, with no mention of `lc init`. A `pyproject.toml` or
`astra.yaml` present but no built environment is the right place,
unconverged — a fresh clone is exactly this, since git carries no
`.venv` — and there `lc init` really is the answer.

---

https://claude.ai/code/session_01DqTza2GqRLZ4yY5ywAusQH

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -593,8 +593,19 @@ only checks that the guard is present. Verified empirically, not assumed.
   the design cares about, which is a reach *outside* the project. Note
   this cannot be narrowed to "everything but `.venv`": Landlock unions
   rights over ancestors and cannot subtract, so a writable project is
-  necessarily a writable `.venv`. `uv sync --exact` at the next
-  convergence is what puts the environment back, not the sandbox.
+  necessarily a writable `.venv`.
+  - **The exec allowlist has to survive that**, and it did not at first:
+    `.venv/bin` was granted as a *directory*, so `cp /usr/bin/git
+    .venv/bin/ && git` ran a host tool the allowlist denies by name.
+    Grants there are per *file* now, taken when the policy is built, so
+    what a run writes afterwards is not runnable. `_exec_set` grants no
+    directory anywhere; the rule is the same one `/usr/bin` has always
+    had, and now it is load-bearing twice over.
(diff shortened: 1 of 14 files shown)


EiffL (commented):
LGTM

PR #175, opened 2026-08-18
Layers 2 and 4: identity, provenance, and `lc materialize`

EiffL:
`lc materialize` executes an ASTRA analysis: it plans the graph, runs each recipe under the layer-5 boundary, and commits every output together with the manifest that describes it. Layer 2 folds in, because the fabric is what needs identity and nothing else did.

Layer table: rows **2** and **4** → ✅.

## Results are versioned in the repository

On the DataLad model — **git carries the pointers and the history, git-annex carries the bytes** — without taking the `datalad` package. Every operation the layer needs is one git or git-annex command, and `uv.lock` is `env_version`'s first term, so 43 transitive packages would stale every output in every project on each boto3 bump.

Compatibility is a commitment rather than a coincidence: `lc init` writes a `datalad.dataset.id`, so a project is a DataLad dataset **from birth** (verified: `datalad status` needs no `--force` adoption step), and the run record is the format `datalad rerun` reads. lc never imports datalad, never requires it, and never parses `.datalad/`.

git stops being optional — a project without version control had nowhere to put a result.

## Identity

```
env_version  = sha256(uv.lock ‖ .python-version ‖ canonical install settings)
code_version = sha256(recipe ‖ canonical decisions ‖ env_version)
```

Both length-framed, so a boundary shift between fields cannot produce one digest from two inputs. The git commit is **recorded with every output and hashed into neither** — a commit must not stale the world; per-output code invalidation is available by declaring source files as ASTRA inputs.

The lock scan **refuses** path/directory/editable dependencies (the lock records where they were, not what was in them, so two syncs can install different code while every hash agrees), **reports** registry packages built from sdist, and is **advisory** on non-default dependency groups.

## Execution

| Module | Role |
|---|---|
| `plan.py` | spec × universes → tasks; sub-analyses flatten to `<analysis>.<output>`, a second nesting level is refused rather than ignored |
| `assets.py` | the output directory, the manifest, and `staleness()` — one predicate, two callers |
| `worker.py` | one output: gates, reset, recipe, hash, manifest. Never raises, never writes git. Also the `python -m` entry point the run record names |
| `materialize.py` | the driver: dirty-tree refusal, Dask, and the serialized save/restore loop |
| `dataset.py` | the one git + git-annex seam, routed through the existing `project._run` |
| `identity.py` | the two hashes and the lock scan |

- **Dask owns the ordering.** Tasks are submitted with their upstream futures as arguments; there is no ready-set loop and no hand-rolled topological sort in the execution path.
- **The driver owns git alone.** Concurrent git operations on one repository race on the index lock — this is `datalad-slurm`'s schedule/finish split.
- **A run leaves the tree exactly as clean as it found it.** `ok` → save, `failed`/`blocked`/never-reported → restore, in a `try/finally`. That is what makes the dirty-tree refusal survivable rather than a trap.
- **`--check` differs from a worker by one input value**, not by logic: `None` for anything already classified as would-run, meaning "this is going to change".

`sandbox/policy.py` gains `recipe_policy()` beside `probe_policy()`, sharing one baseline: a recipe may write its own output directory and nothing else under `results/`. Reads are unchanged — a carve-out there is something Landlock cannot express at all.

## Three things running it turned up that the design did not have

- **`.gitattributes` needs a leading `* annex.largefiles=nothing`.** `git annex add` annexes whatever it is handed, so without it the documented save turns `src/fit.py` into a read-only symlink into the object store and the next edit fails with `EACCES`.
- **`git annex add -A` is not a thing.** `git add` needs `-A`; git-annex rejects it. A test extracts the line from `data/README.md` and runs it, because that spelling was wrong once.
- **`git check-ignore` must be asked with a trailing slash and `--no-index`**, or the `results/*` an older scaffold wrote — which ignores the directory's *contents*, not its name — goes unseen, and a materialize reports success while committing nothing.

## Dependencies

`git-annex` (0 transitive; wheel-only, so it sets the CLI's install floor — see the risk note in the plan) and `distributed`. **`datalad` is a dev-group dependency only**: the run record's format fails *silently* (its parser returns nothing on a mismatch and `rerun` then exits 0), so the suite asserts through datalad's own parser **and** runs a real `datalad rerun`, rather than golden-testing our own JSON.

## Verification

```
348 passed          ruff: clean          mypy: clean
```

End-to-end from a clean `uv pip install .` into a throwaway venv (not the global tool env): `lc init` → commit → `--check` → `materialize` → idempotent re-run → decision change cascades to dependents → old bytes recovered from an old commit. Landlock enforced and recorded in `hermeticity`, results annexed, manifests in plain git, `datalad status` clean, one `[DATALAD RUNCMD]` commit per output.

Also pinned by tests, against a real annex and the real boundary: a recipe cannot write into a sibling output (mutation-checked against `Unavailable()`), a recipe *can* read an annexed input through its symlink, a failing recipe commits nothing and leaves `git status` empty, and `python -m lightcone.engine.worker` imports neither click nor rich.

https://claude.ai/code/session_01GaahDTd6djz9PCm1Rh6SP4

### DIFF
--- a/src/lightcone/cli/commands.py
+++ b/src/lightcone/cli/commands.py
@@ -23,8 +23,12 @@ logger = logging.getLogger(__name__)
 
 @functools.cache
 def _console() -> Console:
-    """The rich console, built on first use to avoid startup cost at each
-    invocation of the cli even when the console is not needed."""
+    """Build the rich console, once and on first use.
+
+    Returns:
+        The console. Deferred so ``lc --help`` and shell completion never
+        pay for it.
+    """
     from rich.console import Console
 
     return Console()
@@ -40,6 +44,17 @@ class _EngineErrorGroup(click.Group):
     """
 
     def invoke(self, ctx: click.Context) -> object:
+        """Run a command, rendering engine errors as clean CLI errors.
+
+        Args:
+            ctx: The click context.
+
+        Returns:
+            Whatever the command returned.
+
+        Raises:
+            click.ClickException: In place of any ``ProjectError``.
+        """
         from lightcone.engine.project import ProjectError
 
         try:
--- a/src/lightcone/cli/commands.py
+++ b/src/lightcone/cli/commands.py
@@ -248,6 +248,11 @@ def materialize(
     if as_json:
         click.echo(json.dumps(report.as_dict(), indent=2))
     else:
+        if report.notes:
+            # click.echo, not rich, for the same reason `lc run` uses it:
+            # these are the boundary's own lines and their remedies are
+            # meant to be pasted, so reflowing them breaks them.
+            click.echo("\n".join(["", *report.notes]), err=True)
         _render_materialize_output(report, root, dry_run=check_only)
 
     if not report.ok or (check_only and not report.up_to_date):
--- a/src/lightcone/cli/commands.py
+++ b/src/lightcone/cli/commands.py
@@ -203,9 +203,8 @@ def materialize(targets: tuple[str, ...], check_only: bool, as_json: bool) -> No
     """Make the analysis's outputs, and commit each one as it lands.
 
     Each output is committed together with its manifest, in a commit that
-    records the command that produced it. Which is why a run refuses to
-    start on a tree with uncommitted changes: it could not otherwise say
-    which code that was.
+    records the command that produced it. The git tree needs to be clean 
+    before the run can start.
 
     Nothing runs that does not need to. An output is remade when its
     recipe, its decisions, the environment, or any of its declared inputs
--- a/src/lightcone/cli/commands.py
+++ b/src/lightcone/cli/commands.py
@@ -194,31 +194,46 @@ def run(command: tuple[str, ...]) -> None:
     ),
 )
 @click.option(
+    "--refresh",
+    is_flag=True,
+    help=(
+        "Also remake outputs that are behind — still what the analysis "
+        "asks for, but made under an earlier environment."
+    ),
+)
+@click.option(
     "--json",
     "as_json",
     is_flag=True,
     help="Emit the report as JSON on stdout.",
 )
-def materialize(targets: tuple[str, ...], check_only: bool, as_json: bool) -> None:
+def materialize(
+    targets: tuple[str, ...], check_only: bool, refresh: bool, as_json: bool
+) -> None:
     """Make the analysis's outputs, and commit each one as it lands.
 
     Each output is committed together with its manifest, in a commit that
     records the command that produced it. The git tree needs to be clean
     before the run can start.
 
-    Nothing runs that does not need to. An output is remade when its
-    recipe, its decisions, the environment, or any of its declared inputs
-    has changed — by content, so a rebuild that comes out byte-identical
-    stops there instead of cascading.
+    An output is remade when the analysis defines it differently than it
+    was made — a changed recipe or decision — or when one of its declared
+    inputs changed. Inputs are compared by content, so a rebuild that
+    comes out byte-identical stops there instead of cascading.
+
+    An output made under an earlier environment is reported as behind and
+    left alone: it is still what the analysis asks for, and the manifest
+    records the environment and the commit that produced it. Pass
+    --refresh to remake those too.
     """
     from lightcone.engine import materialize as engine
     from lightcone.engine.project import current_project
 
     root = current_project()
     if check_only:
-        report = engine.check(root, targets)
+        report = engine.check(root, targets, refresh=refresh)
     else:
-        report = engine.materialize(root, targets)
+        report = engine.materialize(root, targets, refresh=refresh)
 
     if as_json:
         click.echo(json.dumps(report.as_dict(), indent=2))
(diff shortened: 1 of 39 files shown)


EiffL:
## Review findings addressed

Nine of the ten are real; each was reproduced first and is now pinned by a test. **359 passed, ruff clean, mypy clean.** Pushed as `76f93f1`.

### The two that made the identity model lie

**5 — a run started even when `.venv` no longer matched `uv.lock`.** `uv run --locked` asserts only that the lock still matches `pyproject.toml`, and workers pass `--no-sync`. Reproduced: the recipe imported `packaging 26.3` under a lock saying `24.2`, uv accepted it silently, and the manifest recorded the *new* lock's `env_version`. `project._env_is_current` is uv's own probe and already existed — the run just had to ask. Refusal in `materialize()`, warning in `check()`.

**2 — the run record named the annex object path.** `_inside` resolved symlinks, and every declared input under `data/` is one. Seen in a real commit from the walkthrough:

```
"inputs": [".git/annex/objects/Kk/jQ/SHA256E-s9--ab139e95…txt/SHA256E-s9--ab139e95…txt"]   # before
"inputs": ["data/catalog.txt"]                                                             # after
```

### The rest

| # | Confirmed as | Fix |
|---|---|---|
| 3 | `lc init --check <new-dir>` inside a repo → `FileNotFoundError` traceback | split `_in_repository` (pure, answers for absent dirs) from `_can_ask_git` (adds `is_dir()`) |
| 4 | `--check` rehashed upstreams, so a content-less clone reported a rebuild a real run called up to date | check mode reads the recorded manifest, same rule as a worker |
| 7 | dropping a declared input never staled the output — `code_version` moves for none of it | `staleness` compares the declared input *set*, with its own `Reason` |
| 8 | `git_sha` read per task while HEAD moves under the run → nondeterministic provenance | read once by the driver, handed down |
| 9 | post-recipe block could raise into Dask and take down every task in flight | guarded; reports `failed` instead |
| 6 | `my_project` failed to recognise itself against uv's `my-project`, refusing the run over the project's own code | PEP 503 on both sides |
| 10 | `method: ""` reported "does not declare this decision"; `method: null` rendered the literal `None` into the shell command *and* `code_version` | presence for the first, unset for the second |

Finding 9 turned up a bonus: the sandbox already stops a recipe removing its own output directory (write is granted *inside* it; unlinking it needs write on `results/`). So the guard is exercised through `Unavailable()`, which is the host where a recipe really can do it.

### Not a finding — 1

The claim was that a dependent could read an upstream mid-`git annex add` and hit ENOENT. **The overlap is real**; instrumented, the dependent's recipe starts 3 ms after the save begins and finishes inside it:

```
0.798  save-start     results/baseline/first
0.801  recipe-start
0.830  recipe-end
1.352  save-end
```

**The consequence is not.** git-annex hard-links content into the object store and then renames the symlink over the file, so the path never stops existing and never holds partial bytes. Measured with 4 threads doing full-content reads across 8 files (24 MB) for the duration of `git annex add`: 448 reads, **0 missing, 0 short, 0 wrong bytes**.

Recorded in `CLAUDE.md` rather than left implicit, because the obvious "fix" — moving the save into the task — is exactly what puts git back in the workers and reintroduces the index-lock race the design exists to avoid.

https://claude.ai/code/session_01GaahDTd6djz9PCm1Rh6SP4

src/lightcone/cli/commands.py
```diff
@@ -161,26 +147,106 @@ def _render_init_output(report: ConvergenceReport, directory: Path, *, dry_run:
...
+@click.option(
+    "--check",
+    "check_only",
+    is_flag=True,
+    help=(
+        "Report what would run and why, without executing or committing "
+        "anything; exit 1 if anything is out of date."
+    ),
+)
+@click.option(
```
EiffL, line 196: why do we need that here?

src/lightcone/engine/sandbox/policy.py
```diff
@@ -164,18 +164,30 @@ def utility(name: str) -> Path | None:
...
+    hostile recipe.)
 
     ``results/`` is granted only if it is already there. Convergence
-    creates it, and a policy that made directories would be a side
-    effect nobody asked a *probe* for.
+    creates it, and a policy that made directories would be a side effect
+    nobody asked for — a policy describes, it does not prepare.
 
     Creates the per-run HOME on disk as a side effect; the caller owns
     removing it (see :func:`~lightcone.engine.sandbox.boundary.scope`).
```
EiffL: please streamline this docstring, we don't need the whole history of how it came to be, just be factual and concise

src/lightcone/engine/templates/files/gitattributes.tmpl
```diff
@@ -0,0 +1,16 @@
...
+# Outputs and declared inputs are content, and content is often large:
+# git-annex holds the bytes, git holds a pointer to them. Manifests are
+# small and have to be readable on a clone that has fetched no annex
+# content at all, so they stay in git proper — as does everything else,
+# including analysis code, which must stay writable rather than become a
+# read-only symlink into the object store.
+* annex.largefiles=nothing
+results/** annex.largefiles=anything
+data/** annex.largefiles=anything
+**/.lightcone-manifest.json annex.largefiles=nothing
```
EiffL, line 14: streamline the comment, make it more concise and to the point

src/lightcone/engine/dataset.py
```diff
@@ -0,0 +1,248 @@
...
+
+def require_git_annex() -> None:
+    """Refuse early when git-annex is not reachable as git reaches it.
+
+    Probed *after* :func:`_put_our_bin_first`, and deliberately by the
+    name ``git`` itself searches for: ``git annex`` is not a builtin, it
+    is ``git`` finding a ``git-annex`` executable on ``PATH``. Asking any
+    other way would answer a question git never asks.
+    """
+    _put_our_bin_first()
```
EiffL: let's move these require statements close to the uv required statement, not in this file

src/lightcone/engine/dataset.py
```diff
@@ -0,0 +1,248 @@
...
+
+    Prepend, never append: a system git-annex winning would make the
+    version recorded in the lock a fiction. Idempotent, and applied to
+    our own environment because that is what child processes inherit.
+    """
+    ours = str(Path(sys.executable).parent)
+    path = os.environ.get("PATH", "")
+    if path.split(os.pathsep)[:1] == [ours]:
+        return
+    os.environ["PATH"] = ours + os.pathsep + path if path else ours
```
EiffL: Is this absolutely necessary? if the tool is called by uv shouldn't we expect that uv will have a PATH where its bin folder is properly prioritized?

src/lightcone/engine/dataset.py
```diff
@@ -0,0 +1,248 @@
...
+
+Every command goes through :func:`~lightcone.engine.project._run`, the
+same seam convergence uses, so there is one place the suite has to
+monkeypatch and every invocation is inspectable.
+
+The trap worth knowing: a plain ``git add`` of a file that
+``.gitattributes`` marks ``annex.largefiles=anything`` does **not** annex
+it — it silently commits the bytes into git. ``git annex add`` has to run
+first, which is why :func:`save` is ordered the way it is.
+"""
```
EiffL, line 17: streamline and make it more factual and concise

src/lightcone/engine/project.py
```diff
@@ -237,6 +241,39 @@ def require_uv() -> None:
...
+    """
+    if _env_is_current(directory):
+        return ""
+    return (
+        f"{directory}: the environment does not match uv.lock — recipes would "
+        "import packages the lock does not describe, and every manifest would "
+        "record an environment that never ran. Converge it with `lc init`."
+    )
+
+
```
EiffL: is this really necessary? shouldn't be just mechanical that a drifted environment just is not possible ?

src/lightcone/engine/project.py
```diff
@@ -262,34 +299,112 @@ def _converge_uv_project(c: _Converger, directory: Path) -> None:
...
-            "materialize until it is one.",
-        )
+def _converge_tracked_dir(
+    c: _Converger, directory: Path, name: str, readme: Callable[[], str]
+) -> None:
+    """A directory the repository tracks, plus the README that makes it exist.
+
+    ``data/`` and ``results/`` are both empty in a fresh project, and git
+    carries no empty directories — so without a README neither survives a
+    clone, and nothing on disk explains what belongs in it.
```
EiffL: don't mention data and results in a general function

EiffL (commented):
made some comments

src/lightcone/cli/commands.py
```diff
@@ -161,26 +162,98 @@ def _render_init_output(report: ConvergenceReport, directory: Path, *, dry_run:
...
+    is_flag=True,
+    help="Emit the report as JSON on stdout.",
+)
+def materialize(targets: tuple[str, ...], check_only: bool, as_json: bool) -> None:
+    """Make the analysis's outputs, and commit each one as it lands.
+
+    Each output is committed together with its manifest, in a commit that
+    records the command that produced it. Which is why a run refuses to
+    start on a tree with uncommitted changes: it could not otherwise say
+    which code that was.
```
EiffL:
```suggestion
    Each output is committed together with its manifest, in a commit that
    records the command that produced it. The git tree needs to be clean 
    before the run can start.
```

PR #178, opened 2026-08-19
The engine is the host's uv tool, not a project dependency

EiffL:
`lc init` pinned `lightcone-cli` into every project's own lock, which is what forced **layer 3** — a launcher that discovers, mode-detects, scrubs `UV_*`, converges, and delegates to the lc inside the project's venv. Remove the pin and there is nothing to delegate to: the `lc` that was invoked is the engine that runs. Layer 3 is now marked removed by decision rather than pending.

This reverses spec §2's engine-in-lock rule. The spec is a reference the rebuild drifts from deliberately; the reversal lands as a Recorded decision in `CLAUDE.md` rather than waiting on a spec rewrite.

## What changes

A scaffolded `pyproject.toml` declares **no dependencies**. A project's resolution is no longer constrained by the engine's own pins (no astra-tools/click/rich/distributed/git-annex in any project lock), and `requires-python` now comes from the interpreter that wrote `.python-version` instead of from lightcone-cli's metadata — one source, so the bound and the pin cannot disagree. The git-annex wheel floor gates installing the tool and nothing else.

## What the pin bought, and where each guarantee went

- **The engine inside `env_version`** — given up deliberately. An lc upgrade now moves nothing; no output ever reads `behind` over an engine release. This completes the reversal that already took `env_version` out of `definition_version`. The engine becomes attestation (`lc_version` in every manifest), not identity.
- **Reruns reconstruct the exact engine** — re-provided in the record itself. `cmd` pins the engine by version through an ephemeral `uv run --no-project --with lightcone-cli==<v>`, keeping the **module spelling** so the worker stays out of `lc --help` and off `$PATH`. A dev build's version is unpublished, so its record is the bare module — which only exists in a checkout whose own interpreter imports it.
- **Driver/worker skew "structurally impossible"** — moot while workers are in-process threads; noted as a layer-7 connect-probe obligation.
- **Layer 6** — the generated image can no longer get its in-image engine from the lock; noted that the Containerfile must install it as an explicit layer and that the version becomes a tag input.

## The worker now owns the project environment

It has to. A rerun checks out the lock but never the `.venv`, and **`uv run --no-sync` against a missing `.venv` silently creates an empty one** (measured, uv 0.12.5) — recipes would have run in it while the manifest recorded the lock's `env_version`. `worker.main()` therefore syncs before anything executes, which also inherits the loud stale-lock refusal the outer `uv run --locked` used to provide. `current_project(synced=False)` is the affordance: the lock is repository content, the `.venv` is not.

A new test reruns on a **fresh clone**, because the in-place rerun passes with or without the sync.

`_engine_path()` is deleted — the suite faked the pin by injecting `src/` onto `PYTHONPATH`, and there is no pin left to fake. `test_the_run_record_is_what_datalad_reads` tightens its `.endswith` to a full-string match so a prefix appearing or vanishing cannot slip past.

## Verification

- `uv run pytest` — 413 passed; `ruff` and `mypy --strict` clean.
- End-to-end by hand: scaffolded a project (no `lightcone-cli` in `uv.lock`), materialized, `lc status`, then `datalad rerun` on a fresh clone with no `.venv` — the worker created it and the output came back byte-identical.
- The pinned shape run literally, from a wheel built off this branch: an ephemeral engine env reproduced the output with a matching `data_version`, and git-annex is present beside that env's interpreter so `put_our_bin_first()` still resolves.

Incidentally fixes a latent bug in `eval.yml`: the eval project's pin resolved a *released* engine from PyPI rather than the branch under test. No workflow change needed.

## Follow-up, not in this PR

The `UV_*` ambient scrub (spec §4 step 3) is worth having and is independent of the pin — it protects `env_version`'s install-settings term, not the delegation that is gone.

https://claude.ai/code/session_01PGfhkvU3Ee2hhHcwAkeidg

### DIFF
--- a/src/lightcone/engine/assets.py
+++ b/src/lightcone/engine/assets.py
@@ -250,6 +250,11 @@ class Manifest:
     git_sha: str
     #: The `origin` URL, or empty when the repository has no remote.
     git_remote: str
+    #: The engine that made it. Attestation, not identity — it is outside
+    #: both hashes, so an lc upgrade neither stales an output nor puts it
+    #: behind. With the engine outside the project's lock too, this and
+    #: the run record are the whole of what says which engine ran. Empty
+    #: for an engine imported from a source tree with no metadata.
     lc_version: str
     #: What the sandbox actually enforced, as the boundary attested it.
     hermeticity: dict[str, Any]
--- a/src/lightcone/engine/assets.py
+++ b/src/lightcone/engine/assets.py
@@ -250,11 +250,8 @@ class Manifest:
     git_sha: str
     #: The `origin` URL, or empty when the repository has no remote.
     git_remote: str
-    #: The engine that made it. Attestation, not identity — it is outside
-    #: both hashes, so an lc upgrade neither stales an output nor puts it
-    #: behind. With the engine outside the project's lock too, this and
-    #: the run record are the whole of what says which engine ran. Empty
-    #: for an engine imported from a source tree with no metadata.
+    #: The engine that made it. Attestation, not identity: outside both
+    #: hashes, so an lc upgrade neither stales an output nor puts it behind.
     lc_version: str
     #: What the sandbox actually enforced, as the boundary attested it.
     hermeticity: dict[str, Any]
--- a/src/lightcone/engine/dataset.py
+++ b/src/lightcone/engine/dataset.py
@@ -18,37 +18,10 @@ asks anyone to run a git-annex command by hand.
 
 from __future__ import annotations
 
-import os
-import sys
 from collections.abc import Iterable
 from pathlib import Path
 
 from lightcone.engine import project
-
-
-def put_our_bin_first() -> None:
-    """Prepend the directory holding our own interpreter to ``PATH``.
-
-    ``git annex`` is dispatched by git searching ``PATH`` for a
-    ``git-annex`` executable, and ``uv tool install lightcone-cli`` links
-    only *our* entry points into ``~/.local/bin`` — the git-annex we
-    resolved as a dependency sits beside the interpreter instead, and this
-    prepend is the only way git finds it. Verified: with only the tool's
-    ``lc`` on ``PATH``, ``git annex version`` fails and ``lc init``
-    refuses.
-
-    Prepend, never append: the annex we resolved is the one the engine was
-    installed with and the one its behaviour is tested against, and a
-    system copy of another vintage winning that race is a difference
-    nothing records. Idempotent, and applied to our own environment
-    because that is what child processes inherit.
-    """
-    ours = str(Path(sys.executable).parent)
-    path = os.environ.get("PATH", "")
-    if path.split(os.pathsep)[:1] == [ours]:
-        return
-    os.environ["PATH"] = ours + os.pathsep + path if path else ours
-
 
 # =============================================================================
 # The repository
@@ -170,7 +143,6 @@ def require_committer(directory: Path) -> None:
     Raises:
         ProjectError: If no committer identity resolves.
     """
-    put_our_bin_first()
     proc = project._run(["git", "var", "GIT_COMMITTER_IDENT"], cwd=directory)
     if proc.returncode != 0:
         raise project.ProjectError(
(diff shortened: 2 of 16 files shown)


src/lightcone/engine/materialize.py
```diff
@@ -584,6 +583,30 @@ def run_record(root: Path, task: Task, dsid: str) -> str:
...
+    Returns:
+        The command line ``datalad rerun`` will hand to a shell.
+    """
+    module = f"python -m lightcone.engine.worker {target}"
+    v = worker.lc_version()
+    if v and "dev" not in v:
+        return f"uv run --no-project --with lightcone-cli=={v} -- {module}"
+    return module
+
+
```
EiffL: Let's not worry about the dev case. move back to inlining the published version only

EiffL (commented):
one comment

src/lightcone/engine/materialize.py
EiffL: Done in 500d8e0, with the follow-up from our discussion folded in: `_worker_cmd` is gone and the record is one inlined shape, always pinned — but instead of leaving dev builds unresolvable, the requirement pins them by source commit (`lightcone-cli @ git+<Repository url>@<sha>`, sha from hatch-vcs's version, URL from `[project.urls]`). Verified end to end against GitHub: a fresh clone's `datalad rerun` fetched the engine at the pinned commit and reproduced the output byte-identically. The e2e tests now execute the shipped record shape against a wheel built from the working tree.

src/lightcone/engine/assets.py
```diff
@@ -250,6 +250,11 @@ class Manifest:
     git_sha: str
     #: The `origin` URL, or empty when the repository has no remote.
     git_remote: str
+    #: The engine that made it. Attestation, not identity — it is outside
+    #: both hashes, so an lc upgrade neither stales an output nor puts it
+    #: behind. With the engine outside the project's lock too, this and
+    #: the run record are the whole of what says which engine ran. Empty
+    #: for an engine imported from a source tree with no metadata.
```
EiffL: make this shorter

src/lightcone/engine/dataset.py
```diff
@@ -32,14 +32,16 @@ def put_our_bin_first() -> None:
...
+    refuses.
+
+    Prepend, never append: the annex we resolved is the one the engine was
+    installed with and the one its behaviour is tested against, and a
+    system copy of another vintage winning that race is a difference
+    nothing records. Idempotent, and applied to our own environment
+    because that is what child processes inherit.
     """
     ours = str(Path(sys.executable).parent)
     path = os.environ.get("PATH", "")
```
EiffL: This is a little bit hacky, is there another way in pyproject.toml to express that we want git-annex to be accessible from the cmd line? like is it possible to say that we need some uv tools?

EiffL (commented):
minor comments

src/lightcone/engine/dataset.py
EiffL: Good instinct — settled in f549f15, and it turned out better than a uv feature: the git-annex wheel (psychoinformatics-de/git-annex-wheel) ships its executables as console-script entry points on one dispatcher (`git_annex:cli`, which `execv`s the real binary with `argv[0]` preserved), so lightcone-cli now re-declares those four entry points verbatim in its own `[project.scripts]`. An installer links the requested package's executables, so a plain `uv tool install lightcone-cli` puts `git-annex`, `git-annex-shell` and the two `git-remote-*` helpers on PATH by construction — verified end to end, duplicate names are byte-identical scripts so the venv-level collision is benign, and a new test pins our declarations against the wheel's own metadata so an upstream rename fails the suite rather than users' installs. There is no pyproject metadata to link a *dependency's* executables (`--with-executables-from` exists but is an install-time flag), and vendoring binaries into platform wheels was rejected as a repack pipeline for what four lines of metadata provide. The prepend stays: it covers lc's own subprocesses in contexts with no linked bin (uvx, dev checkouts).

EiffL (commented):
LGTM

PR #180, opened 2026-08-20
Layer 6: the container hatch — OCI runtimes as the exec boundary

EiffL:
## What this is

The container hatch, landed whole: declaration, identity, `lc build`, containerized `lc run` and `lc materialize`, manifest v2, and the `lc status` headers. The design decisions were taken interactively and are recorded in CLAUDE.md; the load-bearing ones:

- **Engine on host.** The container is the *recipe's* execution world, never the engine's — driver, git, annex, dask and classification all stay the invoked `lc`. This deviates from spec v6.1's full-stack-in-image rule, recorded: v6.1's reason (the host-sync deadlock) is solved by converging `.lightcone/venv` *inside* the image, and the spec's own Perlmutter row ("recipe wrap, step 3 only") is this exact shape. No delegation machinery, no git through a `--userns=keep-id` bind mount, no engine version in the image tag.
- **The user never sees a Containerfile.** The whole surface is Modal-shaped TOML — `base` / `apt-install` / `run-commands` / `env`, a closed, fully-hashed key set (`pip_install` deliberately has no equivalent; the lock owns Python). The render exists only in a transient build context; `LABEL io.lightcone.image` keeps the image self-describing.
- **The dataset is the image store; runtime stores are caches.** `lc build` saves the image as a docker-archive at `.datalad/environments/<tag>/image` (the `datalad containers-add` layout), annexed and committed — exact bytes to every clone via `git annex get`, no registry. Execution pins the id read from the archive itself. A dropped archive never substitutes; a rebuild is a new commit under a new id.
- **The mount table is the mechanism.** One `OCIBackend` (podman recommended, docker accepted — data-parameterized spellings, one shape) maps the same `Policy` to `:ro`/`:rw` mounts with `--network none`; the attestation (`fs: declared`, `network: denied`) is derived flag-for-flag. No in-container Landlock, no seccomp probe: the engine container never gets the tree `:rw`, so mounts alone express the whole policy. The archive format is chosen so apptainer/singularity become thin layer-7 backends.
- **The record stays runtime-neutral through the worker.** Verified against datalad-container's source: templates expand at *record* time and `rerun` executes the literal string, so a bare-recipe `cmd` was rejected. The containerized record adds `extra_inputs: [<archive>]`, which stock `datalad rerun` fetches through the annex before executing.

## Notable traps found on the way

- git-annex routes dotfiles to git regardless of `annex.largefiles` — without a per-add `annex.dotfiles=true` the archive lands as a full git blob, silently. Pinned both ways in `test_dataset.py`.
- A standalone `chmod -R /opt` layer copy-on-writes the whole interpreter into every archive, doubling it; the chmod rides inside the layers that write `/opt`.
- `boundary.env_argv`'s host-resolved `env` binary need not exist in the image (NixOS), so world-backends apply the overlay natively via `--env` (`contains_prefix`, declared on every backend).
- The dirty check now runs *before* the converge in materialize: the image commit would otherwise sweep the user's staged edits.

## Testing

- `tests/test_image.py`, `test_sandbox_oci.py` — pure (declaration, document, render ordering, tag sensitivity, mounts, attestation, seam composition).
- `tests/test_container.py` — the lifecycle against a stubbed `project._run` (three `ensure_image` strictnesses, refusals mutation-checked, build/save/commit argv).
- `tests/test_container_smoke.py` — real podman end to end, gated (`skipif` + `LC_CONTAINER_TESTS_REQUIRED=1` on Linux CI + two guard tests): build → id/label verification → probe (refusal → success → mount denial → loopback/egress) → full materialize with manifest assertions → **`datalad rerun` on a bytes-free clone fetching the archive through the annex and reproducing the output**.
- 483 tests pass; ruff and mypy clean; `lc --help` still 65 ms.

https://claude.ai/code/session_012oVCfc4eZdyKqT1zY95Xi2

### DIFF
--- a/src/lightcone/cli/commands.py
+++ b/src/lightcone/cli/commands.py
@@ -182,6 +182,14 @@ def run(command: tuple[str, ...]) -> None:
 # =============================================================================
 
 
+def _announce_build(tag: str) -> None:
+    """Say a minutes-long image build is coming — one wording, both verbs."""
+    _console().print(
+        f"building [bold]{tag}[/bold] if nothing refuses first — the first build "
+        "after an environment change can take minutes"
+    )
+
+
 @main.command()
 @click.option(
     "--json",
@@ -203,7 +211,7 @@ def build(as_json: bool) -> None:
     from lightcone.engine.project import current_project
 
     root = current_project()
-    state, tag = engine_container.image_state(root)
+    state, tag, _ = engine_container.image_state(root)
     if state == "direct":
         if as_json:
             click.echo(json.dumps({"mode": "direct"}))
@@ -214,10 +222,7 @@ def build(as_json: bool) -> None:
             )
         return
     if state == "absent" and not as_json:
-        _console().print(
-            f"building [bold]{tag}[/bold] — first build after an environment change; "
-            "this can take minutes"
-        )
+        _announce_build(tag)
     runtime, action = engine_container.build(root)
     if as_json:
         click.echo(
@@ -296,12 +301,9 @@ def materialize(
         # so before handing over. Conditional mood, deliberately: the
         # engine's own refusals (a dirty tree, an invalid spec) come
         # first and cost no build, so this must promise nothing.
-        state, tag = engine_container.image_state(root)
+        state, tag, _ = engine_container.image_state(root)
         if state == "absent":
-            _console().print(
-                f"this run will build [bold]{tag}[/bold] — first run after an "
-                "environment change; this can take minutes"
-            )
+            _announce_build(tag)
     if check_only:
         report = engine.check(root, targets, refresh=refresh)
     else:
@@ -356,8 +358,7 @@ def status(as_json: bool) -> None:
         described = {
             "present": "built",
             "absent": "needs build — run `lc build`",
-            "unfetched": f"content not fetched — git annex get "
-            f".datalad/environments/{tag}/image",
+            "unfetched": f"content not fetched — git annex get {report.image['archive']}",
         }[state]
         lines.append(f"  image:   {tag} — {described}")
     lines.append(f"  sandbox: {report.sandbox}")
--- a/src/lightcone/cli/commands.py
+++ b/src/lightcone/cli/commands.py
@@ -178,6 +178,65 @@ def run(command: tuple[str, ...]) -> None:
 
 
 # =============================================================================
+# lc build
+# =============================================================================
+
+
+@main.command()
+@click.option(
+    "--json",
+    "as_json",
+    is_flag=True,
+    help="Emit the result as JSON on stdout.",
+)
+def build(as_json: bool) -> None:
+    """Build the project's system-layer image, and commit it.
+
+    Containerized projects only — a project containerizes by declaring
+    [tool.lightcone.image] in pyproject.toml. The image is saved into the
+    repository (.datalad/environments/) as versioned content, so clones
+    obtain the exact bytes with `git annex get` instead of rebuilding.
+    Idempotent: an image that is already built and committed is left
+    alone.
+    """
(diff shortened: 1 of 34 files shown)


src/lightcone/cli/commands.py
```diff
@@ -177,6 +177,70 @@ def run(command: tuple[str, ...]) -> None:
...
+# =============================================================================
+
+
+def _announce_build(tag: str) -> None:
+    """Say a minutes-long image build is coming — one wording, both verbs."""
+    _console().print(
+        f"building [bold]{tag}[/bold] if nothing refuses first — the first build "
+        "after an environment change can take minutes"
+    )
+
```
EiffL: don't make a function jsut for this

src/lightcone/cli/commands.py
```diff
@@ -226,10 +290,20 @@ def materialize(
...
     root = current_project()
+    if not check_only and not as_json:
+        # The engine never prints, and the build it may be about to run
+        # can take minutes — so the one place that owns the console says
+        # so before handing over. Conditional mood, deliberately: the
+        # engine's own refusals (a dirty tree, an invalid spec) come
+        # first and cost no build, so this must promise nothing.
+        state, tag, _ = engine_container.image_state(root)
+        if state == "absent":
+            _announce_build(tag)
```
EiffL: Print a short inline message here, like 'image absent, container rebuild necessary' or something to that effect.

EiffL: hold on, is this compatible with the idea that materialize doesn't build?

src/lightcone/cli/commands.py
```diff
@@ -278,9 +352,20 @@ def status(as_json: bool) -> None:
         click.echo(json.dumps(report.as_dict(), indent=2))
         return
 
+    lines = [f"  mode:    {report.mode}"]
+    if report.image is not None:
+        tag, state = report.image["tag"], report.image["state"]
+        described = {
+            "present": "built",
+            "absent": "needs build — run `lc build`",
+            "unfetched": f"content not fetched — git annex get {report.image['archive']}",
```
EiffL: I haven't read the rest of the implementation yet, but quick note that fetching the image should be done by the lightcone cli directly, not asking the human/agent to run a distinct git command.

src/lightcone/engine/assets.py
```diff
@@ -27,7 +27,7 @@
 from lightcone.engine.project import ProjectError
 
 MANIFEST_FILENAME = ".lightcone-manifest.json"
-SCHEMA_VERSION = 1
+SCHEMA_VERSION = 2
```
EiffL: we don't need to bump the schema version here, we haven't released the project yet

src/lightcone/engine/image.py
```diff
@@ -0,0 +1,371 @@
...
+
+#: The default base image. Digest-pinned to the multi-arch manifest list,
+#: so the reference is one string on every architecture and the image
+#: stays a pure function of the repository plus the engine. An engine
+#: release that bumps this moves both the tag and ``env_version`` for
+#: containerized projects — the system layer genuinely changed.
+DEFAULT_BASE = (
+    "docker.io/library/debian:bookworm-slim"
+    "@sha256:abd67ffcfa541b485a3dff59865ab629aa048a6c613e639d36e7456b0b229241"
+)
```
EiffL, line 64: is this image the same default as what modal uses? does it already have python in it?

src/lightcone/engine/image.py
EiffL, line 1: there are a number of custom functions to edit/build the containerfile, are they not well established/standard libraries for building a containerfile programmatically instead of writing our own code?

EiffL (commented):
some small comments

PR #181, opened 2026-08-20
Layer 7: venues — SLURM in-allocation execution and podman-hpc

EiffL:
The first HPC venue, NERSC-focused, in two commits.

## The SLURM venue (7a)

`lc materialize` inside a SLURM allocation now spans every node it was granted: `cluster_for_run()` detects `SLURM_JOB_ID`, binds the Dask scheduler in the driver process to `SLURMD_NODENAME`, and launches one worker per node with a single `srun` on the driver's own interpreter — so driver and workers are the identical installation off the shared filesystem, which is all a worker needs (`lightcone.engine` importable at the driver's version; no git, no annex). Detected, never configured: the allocation *is* the resource declaration, the same shape as mode derivation. No new dependency.

```
salloc -N 4 -C cpu -q regular -t 2:00:00
lc materialize            # runs across all 4 nodes

# or, fire-and-forget from a login node:
sbatch -N 4 -C cpu -q regular -t 2:00:00 --wrap 'lc materialize'
```

A NERSC login node refuses `lc materialize` first thing, naming both commands above; `--check`, `status` and `lc run` work anywhere. A dead srun reports its own exit code; teardown retires workers so srun ends silently. A dask-jobqueue submission venue, if ever wanted, is one more branch in the same ladder plus the config table it genuinely needs.

The suite fakes the *host*, never the code: SLURM env vars plus a bash stub for srun drive the real graph through the real detection → bind → launch → teardown path in CI, with real worker processes. A new `LocalCluster(processes=True)` test pins the cross-process pickling contract the recorded decision asked layer 7 to re-verify.

## podman-hpc (7b)

A containerized project now runs on NERSC: podman-hpc lands as a spelling inside the existing `OCIBackend` — no new shape — plus one step, `podman-hpc migrate <id>`, placed outside the load branch (a fresh build never loads, yet compute nodes only see migrated images) and run driver-side once for all nodes. Detection prefers the wrapper over bare podman (whose node-local store is exactly what compute nodes cannot see); it is build-capable, making a login node the source of matching-arch archives. A foreign-arch archive refuses before the load, naming both arches and the fix. apptainer/singularity stay deferred; CLAUDE.md records the design headroom they need (Runtime stays facts; honest `network: allowed` for a runtime that cannot deny; the exit-125 note is a podman/docker-family fact).

## Verification

- 519 tests pass locally (`uv run pytest`), ruff and mypy clean.
- The container smoke suite gains podman-hpc as a parametrization that fires only where the site tool exists — running it on a Perlmutter login node is the spike vehicle; the checklist (srun `--overlap`, `--network none` on compute, `--module` mounts vs `fs: declared`, migrate-by-id, Landlock in the SLES LSM list, Lustre annex timings) is recorded in CLAUDE.md.

Deviation from the approved plan, recorded: the plan called for two PRs; repo convention is one PR per layer (#175, #180), so both parts land here as separate commits.

https://claude.ai/code/session_012oVCfc4eZdyKqT1zY95Xi2

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -1578,11 +1578,18 @@ count refuses naming the variable (`venue._int_env`), and a
 the bind rather than the raw `socket.gaierror` (distributed wraps it in
 a `RuntimeError`, so the constructor catches both).
 
-**The login guard is materialize-scoped and comes first.**
-`venue.require_compute_node()` refuses iff `NERSC_HOST` is set and
-`SLURM_JOB_ID` is not (NERSC_HOST is set on compute nodes too; the
-allocation is what distinguishes them), naming copy-pasteable `salloc`
-and `sbatch --wrap 'lc materialize'` commands. It is the first line of
+**The login guard is materialize-scoped, table-driven, and comes
+first.** `venue.require_compute_node()` refuses iff a known center's
+marker is in the environment and `SLURM_JOB_ID` is not (a marker is set
+on compute nodes too; the allocation is what distinguishes them),
+naming that center's copy-pasteable `salloc` and `sbatch --wrap 'lc
+materialize'` commands. The centers live in `venue._SITES` — one row
+each: name, marker variable, and the center's own allocation spellings,
+**verified against the center's documentation, never guessed** (the
+remedies rule). NERSC is the seeded row; supporting another center is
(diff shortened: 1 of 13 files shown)


EiffL (commented):
LGTM

PR #182, opened 2026-08-20
Layer 8: the publication view — the RO-Crate converged, integrity as a status fact

EiffL:
## Summary

The last layer of the rebuild table, rescoped during planning from the spec's "`lc verify`, WRROC export" to two mechanisms with **no new verbs**:

**The project is the crate.** `ro-crate-metadata.json` lives at the project root as a derived artifact, converged by `lc materialize` the way `uv.lock` is: rendered from repository state, string-compared, committed alone only on a difference. Deposit is `git archive`/`datalad export-archive` on a repository that is already an RO-Crate — nothing copied, no bundle, no `lc export`. Maintenance is derived, never configured: a declared `[project].license` turns it on (RO-Crate requires one; materialize must not refuse science over a missing key, and inventing CC-BY-4.0 — the pre-rebuild default — asserts terms over someone's data). Absence is one report line.

**Integrity is a status fact, not a verb.** The hole layer 4 recorded — a hand-edited-and-committed output reads `current` forever because a skip returns the recorded digest (the agent-forged-file scenario) — is closed by history, not hashing: every output is committed, so a hand edit *requires* a commit, and `dataset.last_writer` names it. `lc status` flags any output whose directory's last commit is not its own run record, with sha, author, date and a `git show` remedy. O(history), answers on a bytes-free clone, still exits 0.

## The crate itself

- `engine/crate.py`, the one new module — a **pure** builder over `plan.build()` and the manifests; git enters only as an injected `writer` callable.
- Targets **Provenance Run Crate 0.5 structurally, not vacuously** (the pre-rebuild exporter passed REQUIRED only because the SHACL shapes are vacuous over an empty target class): outputs sharing a manifest `git_sha` are one run — the driver's single HEAD read makes run identity free — giving `OrganizeAction` + workflow-level `CreateAction` per run, `ControlAction` per execution, `HowToStep` per output id.
- The workflow-run `@context` makes `containerImage`/`sha256`/`ContainerImage` real vocabulary (the old exporter emitted them as undefined terms JSON-LD silently drops). The committed image archive is one entity: `File` + `ContainerImage`, config-blob id and payload together.
- The manifest stays canonical: in the crate as a `File`, `subjectOf` its `Dataset`; `env_version`/`definition_version`/`hermeticity` are deliberately not transliterated.
- Deterministic by construction: sorted iteration, `sort_keys`, no clock (`datePublished` = newest manifest `finished_at`, overriding rocrate's stamp-now default), no `mimetypes` (it reads `/etc/mime.types`, which would make the render host-dependent).
- The crate's `Person` is the author of each output's **saving** commit via `last_writer` — never the manifest's `git_sha`, which is the commit the run *started* at.

## Validation

Official `rocrate-validator` against the Provenance profile, on a real materialized project: **REQUIRED clean**; RECOMMENDED failures pinned as a *set* (`test_crate_smoke.py::_FLOOR`) of checks the crate cannot truthfully satisfy (the workflow's id is its in-crate path and the shape wants `^http`; an annex-stored image has no registry; lc knows no publisher/affiliation). A new failure is a regression; a disappearing one is the floor to shrink. Gated with `LC_CRATE_TESTS_REQUIRED=1` on every CI runner — the validator is a dev dep, so none may skip.

## Also in here

- The worker stamps `started_at`/`finished_at` (millisecond ISO 8601 — crate consumers parse `endTime` with at most three fractional digits). Attestation like `lc_version`; `classify` never reads them; `SCHEMA_VERSION` stays 1.
- `materialize.run_subject` is now the one spelling of the run record's subject, shared by `run_record` (composer) and the foreign-write check (comparator) — byte-identical to the previous inline string, and the datalad parser + real `datalad rerun` tests all pass.
- Recorded rejections, with reasons in CLAUDE.md: no `lc verify` (O(bytes), blind on unfetched outputs, no lifecycle moment enforcing it; `git annex fsck` owns object corruption), no `lc export`, and the honest residues stated (a forged run-record subject defeats the history check; a manifest lc itself mis-recorded is the one case only a rehash would catch; the rerun entry point does not regenerate the crate).
- One trap the suite found live: forging a result **in place** dirties its byte-identical sibling through the shared thin annex object — the recorded thin hazard demonstrating itself. Tests unlink first; CLAUDE.md records it.

## Test plan

- 556 passed (31 new), ruff and mypy clean, `LC_CRATE_TESTS_REQUIRED=1` locally.
- New: `test_crate.py` (pure, 14 tests, no git — determinism, run grouping, the Provenance edges, ContainerImage, Person, license spellings), `test_crate_smoke.py` (real materialize → real validator, two guard tests), `last_writer` against real repos incl. the enclosing-repository case, the foreign-write fact end-to-end incl. a bytes-free clone, crate convergence (created once / idempotent re-run commits nothing / license added later / license removed keeps the file / spec-dropped output excluded and named).
- Manual e2e: `lc init` → license → materialize (crate lands in its own trailing commit) → re-materialize (no commit) → validator (0 REQUIRED) → forge + commit → `lc status` names the foreign commit.

https://claude.ai/code/session_01Q9ikHqeMA9tTX1Jrx5DdNL

### DIFF
--- a/src/lightcone/cli/commands.py
+++ b/src/lightcone/cli/commands.py
@@ -360,15 +360,17 @@ def status(as_json: bool) -> None:
     lines.append("")
     marks = {"current": "[dim]·[/dim]", "behind": "[cyan]·[/cyan]", "stale": "[yellow]![/yellow]"}
     width = max((len(o.output) for o in report.outputs), default=0)
-    lines += [
+    for o in report.outputs:
         # The commit gets a column of its own, for every state and not only
         # the interesting ones: "which code made this" is the question the
         # verb exists to answer, and it has an answer for a current output
         # too.
-        f"  {marks[o.status]} {o.status:<8} {o.output:<{width}}  "
-        f"{o.git_sha[:7] or '—':<7}" + (f"  [dim]{o.why}[/dim]" if o.why else "")
-        for o in report.outputs
-    ]
+        lines.append(
+            f"  {marks[o.status]} {o.status:<8} {o.output:<{width}}  "
+            f"{o.git_sha[:7] or '—':<7}" + (f"  [dim]{o.why}[/dim]" if o.why else "")
+        )
+        if o.foreign_write:
+            lines.append(f"      [yellow]! foreign write:[/yellow] {o.foreign_write}")
     lines += [f"  [yellow]![/yellow] {warning}" for warning in report.warnings]
 
     counts = report.counts
--- a/src/lightcone/cli/commands.py
+++ b/src/lightcone/cli/commands.py
@@ -365,12 +365,13 @@ def status(as_json: bool) -> None:
         # the interesting ones: "which code made this" is the question the
         # verb exists to answer, and it has an answer for a current output
         # too.
+        # A foreign write arrives as an ordinary stale with its message in
+        # `why`, so the one rendering path covers it; the dedicated field
+        # exists for machine consumers of `--json`.
         lines.append(
             f"  {marks[o.status]} {o.status:<8} {o.output:<{width}}  "
             f"{o.git_sha[:7] or '—':<7}" + (f"  [dim]{o.why}[/dim]" if o.why else "")
         )
-        if o.foreign_write:
-            lines.append(f"      [yellow]! foreign write:[/yellow] {o.foreign_write}")
     lines += [f"  [yellow]![/yellow] {warning}" for warning in report.warnings]
 
     counts = report.counts
--- a/src/lightcone/cli/commands.py
+++ b/src/lightcone/cli/commands.py
@@ -360,18 +360,17 @@ def status(as_json: bool) -> None:
     lines.append("")
     marks = {"current": "[dim]·[/dim]", "behind": "[cyan]·[/cyan]", "stale": "[yellow]![/yellow]"}
     width = max((len(o.output) for o in report.outputs), default=0)
-    for o in report.outputs:
-        # The commit gets a column of its own, for every state and not only
-        # the interesting ones: "which code made this" is the question the
-        # verb exists to answer, and it has an answer for a current output
-        # too.
-        # A foreign write arrives as an ordinary stale with its message in
-        # `why`, so the one rendering path covers it; the dedicated field
-        # exists for machine consumers of `--json`.
-        lines.append(
-            f"  {marks[o.status]} {o.status:<8} {o.output:<{width}}  "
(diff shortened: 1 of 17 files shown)


src/lightcone/cli/commands.py
```diff
@@ -360,15 +360,17 @@ def status(as_json: bool) -> None:
     lines.append("")
     marks = {"current": "[dim]·[/dim]", "behind": "[cyan]·[/cyan]", "stale": "[yellow]![/yellow]"}
     width = max((len(o.output) for o in report.outputs), default=0)
-    lines += [
+    for o in report.outputs:
```
EiffL: does this also transpires in the json output? an output with a foreign write should be considered stale/corrupted also.

src/lightcone/engine/materialize.py
```diff
@@ -718,13 +803,24 @@ def run_record(root: Path, task: Task, dsid: str, runtime: container.Runtime) ->
...
     return (
-        f"[DATALAD RUNCMD] {task.output_id} [{task.universe_id}]\n\n"
+        f"{run_subject(task.key)}\n\n"
         "=== Do not change lines below ===\n"
         f"{body}\n"
         "^^^ Do not change lines above ^^^"
     )
 
 
+def run_subject(key: Key) -> str:
```
EiffL: can you call this function with a name specific to datalad, otherwise we are not sure what it's going to be

EiffL (commented):
minor comments

src/lightcone/cli/commands.py
EiffL: Addressed in 6a97f14, and it went further than the JSON question: a foreign write now classifies **stale everywhere**, from the one shared walk — `lc status` reports it stale (with `foreign_write` kept as a dedicated JSON field so machine consumers can tell it from any other stale without parsing prose), `--check` plans the remake and exits 1, and the next `lc materialize` performs it. The driver answers the history question and hands each worker a `foreign` flag, since workers have no git — the HEAD discipline on a third value. The separate CLI line is gone: it arrives as an ordinary stale with the message in `why`, one rendering path.

src/lightcone/engine/materialize.py
EiffL: Renamed to `datalad_run_subject` in 6a97f14, with the docstring stating whose format it is: datalad's, matched by `rerun`'s regex, not a spelling of ours to adjust. The composer (`run_record`) and the comparator (the foreign-write check) both go through it, and the byte-for-byte subject is pinned by the existing tests that parse the record with datalad's own `get_run_info` and run a real `datalad rerun`.

EiffL (commented):
LGTM

PR #193, opened 2026-08-21
Converge the annex plumbing: required=true always, pinned where PATH cannot serve

EiffL:
## The problem (measured)

`git annex init` wires the repository to a bare `git-annex` — the filter drivers in `.git/config` and four hooks — resolved from whatever PATH the *researcher's* git runs under. lc itself always has git-annex (it bundles one), but the shell the researcher types `git add` into may hold nothing of the sort: an engine invoked through `uvx` puts nothing on PATH. In that state, a plain `git add` of annexed content prints `git-annex: command not found`, **exits 0, and stages the raw bytes into git history** — a 2 GB dataset lands in git proper, silently, on every clone forever. This failure mode is recorded as a measured test in this PR.

## The fix: convergence owns the plumbing

`lc init` gains an `annex-plumbing` convergence item, right after `git annex init`:

- **`filter.annex.required = true`, always.** The silent exit-0 corruption becomes git's own hard refusal (`fatal: ... clean filter 'annex' failed`, nothing staged), with `lc init` as the remedy — quoted verbatim from a real run in the troubleshooting doc.
- **Where the researcher's shell resolves git-annex, stock plumbing is kept untouched.** The probe is `ambient_annex()`: PATH minus the engine's own bin directory, because launchers front that directory on lc's environment and a naive `which` from inside lc always finds the bundled copy. A stock spelling that works outlives any prunable engine path.
- **Where it doesn't, the plumbing is pinned** to the engine's bundled git-annex by absolute path: the three filter drivers (`process`, plus the `smudge`/`clean` fallback pair so both routes resolve the same binary) and the four hooks. Measured: with the filter pinned, `git add` stages a pointer and `commit` runs the pinned pre-commit hook with no git-annex on PATH at all.
- **Asymmetric repair.** A pin that still runs is never rewritten — even when PATH could now serve, or it is not this engine's copy — because both work and rewriting would flip-flop between a tool install and a uvx cache. Only a broken pin is repaired: toward stock when the shell resolves git-annex, else to the current engine's copy. `lc init --check` reports drift without writing.
- **User hooks are never touched.** Ownership is gated on git-annex's own `automatically configured by git-annex` marker; the hooks directory is asked of `git rev-parse --git-path hooks`, so linked worktrees and `core.hooksPath` behave.

`require_git_annex()` stays a PATH check, unchanged: it gates lc's *own* `git annex` subprocesses, which dispatch from lc's process PATH — a different question from the researcher's shell, and recorded as such in CLAUDE.md.

## Verification

- Full suite green (608 passed, run twice), `ruff check` clean, `mypy src/` strict clean, `zensical build` clean.
- New unit coverage (host-independent, fake tools): fresh pin, stock-plus-required, idempotent re-run, dead pin repaired to engine, dead pin repaired to stock, working pin surviving ambient git-annex appearing, `--check` flags drift and writes nothing, user hook untouched, probe regressions (the `sys.executable` symlink resolve; a tool install's `~/.local/bin/git-annex` symlink counting as genuinely ambient).
- New real-annex integration tests: pointer staged and pinned hook run with git-annex absent from PATH; `required=true` alone failing hard with nothing staged; the exit-0 raw-bytes hazard pinned as a measured fact; stock spellings pinned against a real `git annex init`.
- Docs: `docs/cli/init.md`, `docs/user/troubleshooting.md`, `docs/api/{project,dataset}.md`, `docs/architecture.md`, CLAUDE.md invariants — every added command block executed before shipping.

Companion change: LightconeResearch/agent-skills#34 carries a skill-side interim snippet doing the same convergence from the outside; once this ships in a release, that snippet reduces to nothing.

https://claude.ai/code/session_01LsttSNGJ8kEMavA27ezW9L

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -561,7 +561,20 @@ instead of silent corruption.
 path into a repository and rewrites no filter driver or hook: how git
 dispatches git-annex stays `PATH` resolution, which is git-annex's own
 design and git-lfs's too (it writes `filter.lfs.required = true` beside
-`PATH`-relative drivers and has never baked a path). Pinning the filter
+`PATH`-relative drivers and has never baked a path).
+
+The rule the neighbours agree on is **route through a filter ⇒ set
+`required`**, and both halves were measured. git-lfs routes and sets
+it. `datalad create` does *neither*: no `required`, and no
+`* filter=annex` in its `.gitattributes` (just `* annex.backend=MD5E`
+and `**/.git* annex.largefiles=nothing`) — so a plain `git add big.bin`
+in a DataLad dataset stages 200 000 raw bytes into git *even with
+git-annex working*, and the safeguard is procedural: use `datalad
+save`. lc made the opposite promise — the researcher types the git they
+already know — so lc owns the failure mode that promise creates. The
+honest cost, also measured: once a project holds committed annexed
+content, the refusal covers `git status`, `git diff` and `git checkout`
+too, not just `git add`. Pinning the filter
 and the four hooks to the engine's bundled executable was implemented
 and reverted before merge: it writes durable state pointing at a
 prunable `uvx` cache, `core.hooksPath` can send the hooks into a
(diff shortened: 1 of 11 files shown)


EiffL (commented):
LGTM

PR #196, opened 2026-08-21
Docs deploy skips pre-releases

EiffL:
Ahead of tagging `v0.5.0rc1`: `release: published` fires for pre-releases too (`released` is the type that skips them), so publishing an rc from the GitHub UI would have run

```
mike deploy --push --update-aliases 0.5.0rc1 latest
mike set-default --push latest
```

— moving the `latest` alias onto the rc and making docs.lightconeresearch.org serve rc docs as the site default. That contradicts the workflow's own stated rule that the site tracks the *released* CLI: PyPI hands an rc only to someone who asks for it by name (`--pre`, or an exact `==0.5.0rc1`), so the site must keep serving the last full release.

The fix is one `if` on the deploy job. `github.event.release` is null on a `workflow_dispatch` run, so the manual redeploy path stays unconditional.

`pypi-publish.yaml` is deliberately untouched. PyPI has no per-upload pre-release flag — it derives one from the PEP 440 version alone. Verified against this branch's build config: a `v0.5.0rc1` tag gives hatch-vcs `0.5.0rc1`, which `pip install lightcone-cli` skips without `--pre`. The GitHub checkbox never enters into it.

Prose updated in the two places that describe the release flow (`docs/contributing/setup.md`, `CLAUDE.md`).

https://claude.ai/code/session_01AujePdgpiCcNnEJ99ruLFu


EiffL (commented):
LGTM

EiffL (commented):
LGTM

PR #199, opened 2026-08-25
One output is one file, named by the spec

EiffL:
An output was a **directory** a recipe filled, so nothing declared what was in it. In a real agent-run project that meant one output holding `chain.npz` *and* `run_config.json`, with the filename contract living only inside three consumer scripts; a recipe exiting 0 having written nothing committing a well-formed digest of an empty directory; and MySTRA guessing which file in a figure's directory was the figure (its own output shows `data-canonical-url=…corner.pdf` next to `src=…corner.png`).

ASTRA now declares the missing half — `Output.format`, astra-spec 0.0.14 / astra-tools 0.2.17 — the counterpart of `Input.source` having always carried a path with an extension.

## The layout

```
results/<universe>/<output_id>.<format>          the output
results/<universe>/.<output_id>.manifest.json    its manifest, in git
```

`format` comes from the spec, so lc composes the filename rather than a recipe choosing it — which makes the whole contents of `results/` a pure function of the spec.

## Scope: a flat analysis, deliberately

This PR handles an analysis without sub-analyses, and nothing else. An earlier revision carried a results tree per sub-analysis, with the guards that needed — a path-collision refusal, an escape refusal, a missing-universe-file refusal, a broadened `**/results/**` routing rule, a path-shaped dirty-tree split, and a multi-root crate walk. All of it was removed: none is required by the change this PR is *for*.

The consequence, stated plainly because it is a real capability removed: **a nested spec is no longer buildable.** ASTRA qualifies an output declared in a sub-analysis as `<analysis>.<output>`, and `output_path` refuses an id carrying a dot — the sidecar is recovered by partitioning the filename on its first dot, so a dotted id would name a manifest two outputs could silently share. On `main` such an output landed in a flat directory named by its dotted id. #202 records what supporting it properly would take.

## What this buys

- **The multiplicity has nowhere to come from.** No directory to compose a second name into, so `savefig(args.out)` writes exactly one file at exactly the declared path.
- **The producer/consumer filename contract is gone** — `{inputs.X}` renders to the upstream's *file*.
- **A consumer that never runs anything knows what every output is**, from `astra.yaml` alone.
- **One digest, not two.** `data_version` drops its `file:` frame, so the manifest, the crate's `sha256` and `sha256sum` agree:
  ```
  manifest data_version: sha256:4f2680c28efa…
  sha256sum            : sha256:4f2680c28efa…
  ```

## Refusals over guesses

All before anything is spent: an output with no `format` (named, all of them at once); an id or format that cannot compose a path; a payload that is not a regular file (`mkdir {output}` used to commit a directory digest); and a manifest the attributes would route into the annex.

## One reversal, recorded

**The sandbox scope widens** from the per-output directory to the directory an output lands in. Not a preference — a mechanism can only grant *creating* a file through the directory that will hold it, and the Landlock shim masks a non-directory grant to `_FILE_ONLY_BITS`. Outputs declared side by side are mutually writable again; `data_version` remains what answers for an output's bytes.

## Breaking, deliberately

Every project restages; old `results/<u>/<o>/` directories are invisible to the new code (`git rm -r results/` and re-materialize — no migration path, per the no-back-compat rule); a spec without `format` on every executable output is refused; multi-file outputs must become separate outputs or a container format; and nested specs are refused (above).

## Verification

`614 passed`, ruff and mypy clean. Beyond the suite, a real project materialized end to end: payloads annexed and manifests git-carried, `tar.gz` yielding `.chain.manifest.json` (first-dot partition, not `.stem`), tree clean after, and a bytes-free clone fetching its inputs and landing on up-to-date with **nothing remade** — which only happens if every manifest stayed readable without content.

## Noted, not fixed

All pre-existing and independent of this change, recorded in `CLAUDE.md`:

- **#200** — a malformed `astra.yaml` tracebacks out of `lc status` and `lc materialize --check` as a raw `ScannerError`, contradicting the invariant it now sits under.
- **#201** — a declared `source:` resolves against the project root wherever it is declared.
- **#202** — nested specs are not buildable, and what deciding it would take (including an astra-tools bug: `validate_universe_file` validates against the *unresolved* tree, so a nested external sub-analysis fails with `[UNKNOWN_ANALYSIS]` at depth 2+).

https://claude.ai/code/session_01PybSgBiP8pjZeWxxCxiAXi


EiffL:
**Claude Code review** — 1 finding(s)

**nit** — evals/prompt.md:33-34

The manifest sidecar was renamed to `.<output_id>.manifest.json`, but two doc references to the old `.lightcone-manifest.json` name were missed: `evals/prompt.md:34` (the very next line was updated to the new `results/baseline/<output_id>.<format>` layout) and `CLAUDE.md:481` in the storage invariants (whose own Recorded decision, added by this PR, documents the new pattern). Both are one-line doc drift with no runtime impact — but CLAUDE.md's own rule is "document only what exists," and the eval prompt is fixture text every future eval agent reads.



---
_Generated by [Claude Code](https://claude.ai/code)_

.github/workflows/eval.yml
EiffL, line 1: it's ok, let's not make things complicated here. remove this spec version fix

cailmdaley:
this is the most Opus 5 PR i've ever seen haha. i got clickbaited by the PR name

docs/api/assets.md
```diff
@@ -16,8 +16,9 @@ Source: `src/lightcone/engine/assets.py`.
 | `Verdict.calls_for_a_remake(refresh=)` | The one place a state becomes an action: `stale` always, `behind` only when asked. |
 | `data_version(path)` | Content hash of a directory or file — computed in the worker, before anything is annexed. |
 | `Versions` | Per-run memo so a shared declared input hashes once, not once per dependent. |
-| `read(dir)` / `write(...)` | The manifest, `.lightcone-manifest.json`. |
-| `output_dir(root, u, o)` | The path, guarded: an id that is not one path component is refused — this guard is what lets the worker's reset stay a whole-directory delete. |
+| `read(sidecar)` / `write(...)` | The manifest, `.<output_id>.manifest.json`. Both take the sidecar's own path, so a caller holding an output path has to say `manifest_path` out loud. |
+| `output_path(analysis_root, u, scope, id, fmt)` | The output's file, guarded: any part that is not a single path component is refused, and so is a format that could not be an extension. |
```
EiffL: why do we still have analysis_root and scope?

src/lightcone/engine/materialize.py
```diff
@@ -513,6 +521,7 @@ def materialize(
         # only maintainer.
         _converge_crate(root, report, full, dsid)
         return report
+    _require_manifests_in_git(root, graph)
```
EiffL: I don't think we need that, can we remove this function and any utilities that are added just for it?

src/lightcone/engine/materialize.py
```diff
@@ -1029,6 +1045,58 @@ def _graph(
     return graph, env_version, full
 
 
+def _require_manifests_in_git(root: Path, graph: Graph) -> None:
```
EiffL: we don't need this, we already have a check that the gitattributes are correct/converged

src/lightcone/engine/materialize.py
```diff
@@ -1029,6 +1045,58 @@ def _graph(
...
+    Args:
+        root: The project root.
+        graph: The run's tasks.
+
+    Raises:
+        ProjectError: If any manifest would be routed to the annex.
+    """
+    if not (paths := sorted({task.manifest_path for task in graph.tasks.values()})):
+        return
+    routed = dataset.largefiles(root, paths)
```
EiffL: is dataset.largefiles used anywhere else? if not, we can remove that function as well.

EiffL (commented):
Left some comments

